PR #2, opened 2023-09-11
Add a GH Actions workflow that builds the dependencies

ricrogz:
This workflow builds all the dependencies required to build sketcher.


schrodinger_build.py
```diff
@@ -0,0 +1,82 @@
+"""
```
cdvonbargen, line 1: Meh, let's leave this out until it's something that works.

src/app/sketcher_app.cpp
```diff
@@ -0,0 +1,11 @@
+#include <QApplication>
```
cdvonbargen, line 1: I posted https://github.com/schrodinger/sketcher/pull/3 with all the skeleton code I put together; let's check that in separately to make this review easier on the(my) eyes

schrodinger_build.py
ricrogz: Deleted.

src/app/sketcher_app.cpp
ricrogz: ok

cdvonbargen:
Closing out this pull request in favor of https://github.com/schrodinger/sketcher/pull/6

PR #4, opened 2023-09-20
Fork from Ricardo's workflow branch

cdvonbargen:


dependencies/dependencies.json
```diff
@@ -0,0 +1,26 @@
...
+        "library": "maeparser",
+        "version": "1.3.1"
+    },
+    {
+        "library": "fmt",
+        "version": "6.1.2"
+    },
+    {
+        "library": "qt6",
+        "version": "6.2.5"
```
cdvonbargen, line 16: should this be 6.2.7?
https://opengrok.schrodinger.com/xref/mmshare/build_tools/library_definitions.json?r=9c596363#309

dependencies/dependencies.json
```diff
@@ -0,0 +1,26 @@
...
+    },
+    {
+        "library": "rdkit",
+        "version": "master"
```
cdvonbargen, line 20: are we on master? Or is a specific hash we should be using?
https://opengrok.schrodinger.com/xref/mmshare/build_tools/library_definitions.json?r=9c596363#318

dependencies/dependencies.json
cdvonbargen: Ok, so it looks like niether 6.2.7 nor 6.2.6 are available?
```
+ curl -L -o qt6-base-6.2.7.tar.xz https://download.qt.io/archive/qt/6.2/6.2.7/submodules/qtbase-everywhere-opensource-src-6.2.7.tar.xz
...
tar: This does not look like a tar archive
xz: (stdin): File format not recognized
...
```

PR #5, opened 2024-01-30
Add fmt and zstd dependencies using FetchContent_Declare

cdvonbargen:
* Linked Case: [SKETCH-2033](https://jira.schrodinger.com/browse/SKETCH-2033)
* Primary Reviewer: @ricrogz 
 
### Description

Branch with an example of using FetchContent_Declare to pull down fmt and zstd. Ricardo originally set this up in a branch by creating a series of bash scripts that build each dependency -- that works, but I recently found FetchContent and wanted to give it a try. The advantage I see is that we'd only have to build the components of each dependency that rdkit_extensions/sketcher actually need. Note the `FIND_PACKAGE_ARGS` argument, which allows the cmake script to bypass pulling down source if the package is already found. 

The posted changes include:
- a simple header/library source/executable to demonstrate the dependencies work
- a simple github action that will cache built dependencies if the workflow succeeds
- a Schrodinger-env specific helper script for us to be able to compile locally with software/lib/

Also, zstd as part of the cmake project has two annoying quirks that aren't present when using targets from an installed zstd:
- it neither uses the zstd:: namespace for targets that are otherwise [installed/exported](https://github.com/facebook/zstd/blob/a7b4dafa0323441f485c2387fd1904da67f386a1/build/cmake/CMakeLists.txt#L190). I manually alias it in that case, but don't need to do this if I run locally with software/lib/. 
- the libzstd_static target doesn't automatically set includes; see https://github.com/facebook/zstd/issues/3758
Not sure what the best way to deal with these, or if this is a sign that I shouldn't be using FetchContent.

Next review: add in boost and boost's unit_test_framework.

### Testing Done

- CI on all 3 platforms
- Ran `>$SCHRODINGER/run schrodinger_build.py && ./build/sketcher_app` locally


ricrogz:
> which allows the cmake script to **pybass** pulling down source

Too much python.

include/schrodinger/sketcher/definitions.h
```diff
@@ -1,17 +1,18 @@
...
 
+#ifdef SKETCHER_BUILD_SHARED_LIBS
 #ifdef _WIN32
 #define API_HELPER_IMPORT __declspec(dllimport)
 #define API_HELPER_EXPORT __declspec(dllexport)
 #else
 #define API_HELPER_IMPORT __attribute__((visibility("default")))
 #define API_HELPER_EXPORT __attribute__((visibility("default")))
 #endif
+#endif
```
ricrogz, line 15:
```suggestion
#define API_HELPER_EXPORT __attribute__((visibility("default")))
#endif
#else
#define API_HELPER_IMPORT
#define API_HELPER_EXPORT
#endif
```

include/schrodinger/sketcher/definitions.h
```diff
@@ -1,17 +1,18 @@
 /**
- * @brief
  * Public definitions for schrodinger::sketcher:: library
  */
 
 #pragma once
 
+#ifdef SKETCHER_BUILD_SHARED_LIBS
```
ricrogz, line 7: Where is this (un)defined?

.github/workflows/sketcher-builder.yml
```diff
@@ -0,0 +1,59 @@
...
+
+      - name: Dependency cache - zstd-1.5.5
+        uses: actions/cache@v4
+        with:
+          path: build/_deps/zstd-*
+          key: zstd-1.5.5-${{ matrix.config.os }}
+
+      - name: Build sketcher
+        run: |
+          cmake . -B build -DBUILD_SHARED_LIBS=OFF -DCMAKE_BUILD_TYPE=Release
```
ricrogz, line 50: Since you are adding the option to the CMakefile (yay, more stuff to maintain), you should also provide a CI to build with shared libs.

cdvonbargen:
> > which allows the cmake script to **pybass** pulling down source
> 
> Too much python.

Spoonerism!

.github/workflows/sketcher-builder.yml
cdvonbargen: Ah, you're right. I'm gonna punt on getting this working for SHARED for the time being and soldier on with setting up static building.

cdvonbargen: Oh, unless you think I should just lock this all down on buiding static only, I'd happily do that.

ricrogz: I think the sketcher makes more sense for users as a static executable than an executable plus a set of shared libraries (easier to handle, harder to have dependency issues). Not sure if you have something else in mind.

cdvonbargen: Cool, switched the CMake over to always build a static library exclusively.

PR #6, opened 2024-03-11
SKETCH-2181. Sketcher/rdkit_extensions build infrastructure; macOS runner works

cdvonbargen:
* Linked Case: SKETCH-2181
* Branch: main
* Primary Reviewer: @ricrogz 
 
### Description
Added a pair of CMakeLIsts.txt and github runner yaml files for building:
- all dependencies as described in external/
- all the rdkit_extensions/sketcher library with a toy set of C++ files exercising the above dependencies

The cmake file in external/ uses `ExternalProject_Add` to pulls dependencies down from github, apply patches if necessary, and configure/build/install the dependency as static libs using cmake itself. The build-external/action.yml action will build one dependency library target in that cmake file if the requested version/os has not been built before, using actions/cache@v4. (Should I also cache on a SHA of the cmake file? a SHA of all of external?)

NOTE: the point of using `ExternalProject_Add` is to hopefully streamline a WASM build since everything is built with CMake.

There are two patches to the dependency libraries:
- RDKit: a patch Ricardo gave me which allows RDKit to build static libraries
- Qt: a patch that updates the QSvg repo dependency to the current name of the qtbase repository -- this goes away when we use a newer version of Qt

The toplevel cmake file builds the rdkit_extensions and sketcher libraries, tests for both libraries, and the sketcher_app executable. The interesting/different bit here is that because we compile static libs in the runners, both libraries require a define that clear out the declspec (LIBNAME_STATIC_DEFINE) which is the only small, necessary departure from our existing definition.h files (which are always used to build shared libraries).

I also have a script that will build this toy system using /software/lib/, but it sadly doesn't work for our installed RDKit builds (see SHARED-10015) so that will have to wait for another day.

List of things I didn't figure out:
- build-external/action.yml's cache path and return value violate DRY but I don't know how to dedupe
- CMAKE_ARGS does shenaigans with `;` forcing me to use `$<SEMICOLON>` in a couple places
- I omitted adding `-fvisibility=hidden` to mac/linux because I got `ld: warning: direct access in function to global weak symbol from file means the weak symbol cannot be overridden at runtime. This was likely caused by different translation units being compiled with different visibility settings.` when building against RDKit. I assume I need to pass that to RDK too, but we can spin that out as a separate ticket.

I can spin those out a cleanup case, but would be happy for any pointers!

### Testing Done
Verified that the macOS runner succeeds; kicked out subtasks for linux, windows, and WASM


cdvonbargen:
Just for reference, I abandoned https://github.com/schrodinger/sketcher/pull/5 which uses FetchContent_Declare for this setup using ExternalProject_Add -- annoyingly, many projects don't adhere to CMake conventions that would make FetchContent_Declare friendly, so I just gave up.

.github/actions/build-external/action.yml
```diff
@@ -0,0 +1,40 @@
...
+        path: build/external/${{ inputs.library }}-${{ inputs.version }}
+        key: ${{ inputs.library }}-${{ inputs.version }}-${{ inputs.os }}
+
+    - name: Configure CMake
+      if: ${{ steps.cache-check.outputs.cache-hit != 'true' }}
+      shell: bash
+      run: |
+        cmake -B build/external/ -S external/ \
+              -DCMAKE_BUILD_TYPE=${{ env.BUILD_TYPE }}
+
```
ricrogz, line 33: I think we only need to do this once (not for each thing we build).

.github/workflows/sketcher-builder.yml
```diff
@@ -0,0 +1,113 @@
...
+        shell: bash
+        run: |
+          {
+              echo 'VERSIONS_JSON<<EOF'
+              cat external/versions.json
+              echo -e '\nEOF'
+          } >> "$GITHUB_ENV"
+
+      - name: CPU count
+        shell: python
```
ricrogz, line 38: If you have python, then you might want to install the "ninja" package and pass "-G Ninja" to CMake. This will make builds faster on windows, as MSBuild cannot build in parallel processes.

external/CMakeLists.txt
```diff
@@ -0,0 +1,115 @@
...
+# Use generator expression to sidestep CMAKE_ARGS removing semicolons
+string(REPLACE ";" "$<SEMICOLON>" BOOST_INCLUDE_LIBS "${BOOST_INCLUDE_LIBS}")
+ExternalProject_Add(
+  boost
+  GIT_REPOSITORY https://github.com/boostorg/boost.git
+  GIT_TAG boost-${BOOST_VERSION}
+  GIT_SHALLOW TRUE
+  CMAKE_ARGS
+    -DCMAKE_INSTALL_PREFIX=${CMAKE_CURRENT_BINARY_DIR}/boost-${BOOST_VERSION}
+    -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=OFF
```
ricrogz, line 68: To allow boost to be found on windows (doesn't affect Linux or Mac, since this is already the default for them):
```suggestion
    -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=OFF -DBOOST_INSTALL_LAYOUT=system
```

ricrogz (approved):

src/app/main.cpp
```diff
@@ -1,11 +1,11 @@
...
-#include "schrodinger/sketcher/sketcher_widget.h"
+#include "schrodinger/sketcher/example_widget.h"
 
 int main(int argc, char** argv)
 {
     QApplication application(argc, argv);
-    schrodinger::sketcher::SketcherWidget sketcher_widget;
-    sketcher_widget.show();
+    schrodinger::sketcher::ExampleWidget widget;
+    widget.show();
```
KevKeating, line 9: Just to make sure that I'm reading things correctly, there's currently no actual Sketcher code in this repo, right?  It just builds the dependencies plus a dummy Hello-World-ish app, right?

cdvonbargen: yup! I'm gonna move all the actual code over as a separate (probably set of) commit(s).

external/CMakeLists.txt
cdvonbargen: Offlined with Ric; I don't need INSTALL_LAYOUT to get this to work on the windows runner. We may need to deal with turning threading off for the WASM build, but kicked that conversation out to the WASM case.

.github/workflows/sketcher-builder.yml
cdvonbargen: Will do.

.github/actions/build-external/action.yml
cdvonbargen: Can do.

.github/workflows/sketcher-builder.yml
```diff
@@ -0,0 +1,120 @@
...
+      - name: Install Ninja
+        run: python -m pip install ninja
+
+      - name: CPU count
+        shell: python
+        run: |
+          import os
+          import multiprocessing
+          with open(os.getenv("GITHUB_ENV"), "a") as fh:
+            fh.write(f"CPU_COUNT={multiprocessing.cpu_count()}")
```
ricrogz, line 48: You can remove CPU_COUNT now. `ninja` will auto detect the number of cores, and use the appropriate number of parallel build processes.

.github/workflows/sketcher-builder.yml
```diff
@@ -0,0 +1,120 @@
...
+                                ${{ steps.fetch-qt.outputs.path }}"
+        run: |
+          cmake -B build -S . -G Ninja \
+                -DCMAKE_BUILD_TYPE=${{ env.BUILD_TYPE }} \
+                -DCMAKE_PREFIX_PATH="${{ env.CMAKE_PREFIX_PATH }}"
+
+      - name: Build
+        run: |
+          cmake --build build --config ${{ env.BUILD_TYPE }} \
+                -j ${{ env.CPU_COUNT }}
```
ricrogz, line 110: This should be removed too

.github/workflows/sketcher-builder.yml
cdvonbargen: Oh, terrific!

ricrogz (approved):
LGTM

PR #7, opened 2024-03-15
SKETCH-2185. Add clang-format runner check

cdvonbargen:
* Linked Case: SKETCH-2185
* Branch: 2024-2
* Primary Reviewer: @ricrogz 
 
### Description
Added a stage that runs clang-format over the C++ source

### Testing Done
Made a whitespace edit and confirmed that the stage fails with a clang-format error

### DIFF
--- a/.github/workflows/sketcher-builder.yml
+++ b/.github/workflows/sketcher-builder.yml
@@ -29,9 +29,7 @@ jobs:
       - uses: actions/checkout@v4
 
       - name: Install Python packages
-        run: |
-          python3 -m pip install ninja
-          python3 -m pip install clang-format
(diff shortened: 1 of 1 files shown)


cdvonbargen:
> What's the point of this check?

To enforce formatting -- I thought this was how we had to handle formatting hooks in github. Is there a better way?

ricrogz:
> To enforce formatting -- I thought this was how we had to handle formatting hooks in github. Is there a better way?

Ok, I wanted to make sure you were not expecting clang-format to actually format the code that gets merged. That requires a different approach (which I'm not sure we can use, due to permission restrictions in the org).

ricrogz (approved):
LGTM

PR #8, opened 2024-04-12
Add jsoncpp and zlib dependencies

cdvonbargen:
* Linked Case: SKETCH-2205
* Primary Reviewer: @ricrogz 
 
### Description
Added zlib and jsoncpp dependencies

### Testing Done
Added example usage of jsoncpp; macOS runner succeeds


ricrogz:
As we commented in standup, our current jsoncpp version is old, and, if I have to trust the benchmarks I've seen, it's bulky and slow. We can try using boost.json, which we either have for free from boost, or at least we can add it with little effort.

cdvonbargen:
Ok I'm gonna discard this review.

PR #10, opened 2024-05-27
Update rdkit and qt, and build zlib

cdvonbargen:
### Description
Changes to the runner workflow:
- created a venv and loaded it into GITHUB_ENV (mac runners updated, see https://github.com/schrodinger/sketcher/actions/runs/9238532228/job/25416692649)
- add `build-output` to the build matrix (to differentiate from WASM builds)
- added a cron schedule to avoid the build going stale

Changes to the example build:
- updated RDKit to the March release (part of the patch isn't needed)
- updated Qt to 6.5.3 (... it seems 6.5.5 isn't tagged in the repo?) which removes the need for the Qt 6.2 patch
- updated Boost to 1.83.0 to address issue with std::unary_function
- include Boost::iostreams and build zlib


external/CMakeLists.txt
```diff
@@ -66,10 +78,18 @@ ExternalProject_Add(
...
     -DCMAKE_INSTALL_PREFIX=${CMAKE_CURRENT_BINARY_DIR}/boost-${BOOST_VERSION}
-    -DCMAKE_BUILD_TYPE=Release -DBUILD_SHARED_LIBS=OFF
-    -DBOOST_INCLUDE_LIBRARIES=${BOOST_INCLUDE_LIBS})
+    -DCMAKE_BUILD_TYPE=Release
+    -DCMAKE_PREFIX_PATH=${CMAKE_CURRENT_BINARY_DIR}/zlib-${ZLIB_VERSION}
+    -DBUILD_SHARED_LIBS=OFF
+    -DBOOST_INCLUDE_LIBRARIES=${BOOST_INCLUDE_LIBS}
+    -DBOOST_RUNTIME_LINK=static
+    -DBOOST_IOSTREAMS_ENABLE_BZIP2=OFF
+    -DBOOST_IOSTREAMS_ENABLE_ZSTD=OFF)
```
ricrogz, line 87: Why? We are providing zstd .

.github/actions/build-external/action.yml
```diff
@@ -22,11 +22,11 @@ runs:
       id: cache-check
       with:
         path: build/external/${{ inputs.library }}-${{ inputs.version }}
-        key: ${{ inputs.library }}-${{ inputs.version }}-${{ inputs.os }}
+        key: ${{ inputs.library }}-${{ inputs.version }}-${{ runner.os }}-${{ matrix.build-output }}
```
ricrogz, line 25: Does this action inherit `matrix` from the parent context? I thought we only got the variables that we explicitly declared as inputs.

.github/actions/build-external/action.yml
```diff
@@ -22,11 +22,11 @@ runs:
...
 
     - name: Build
       if: ${{ steps.cache-check.outputs.cache-hit != 'true' }}
       shell: bash
       run: |
         cmake --build build/external/ --target ${{ inputs.library }} \
-              --config ${{ env.BUILD_TYPE }}
\ No newline at end of file
+              --config ${{ env.BUILD_CONFIG }}
```
ricrogz, line 32: Does it also inherit `env` ?

external/versions.json
```diff
@@ -1,7 +1,9 @@
 {
-    "boost": "1.81.0",
+    "boost": "1.83.0",
     "fmt": "10.1.0",
-    "qt": "6.2.7",
-    "rdkit": "2023.09.4",
+    "qt": "6.5.3",
```
ricrogz, line 4: Re: qt 6.5.5 not being tagged: I guess it's probably in the commercial-only stage. It'll probably become community-accessible when there's another version that supersedes it?comm

.github/workflows/sketcher-builder.yml
```diff
@@ -45,69 +48,78 @@ jobs:
...
-        run: ctest --build-config ${{ env.BUILD_TYPE }} --output-on-failure
+        run: ctest --build-config ${{ env.BUILD_CONFIG }} --output-on-failure
 
       - name: Download sketcher executable
         uses: actions/upload-artifact@v4
         with:
-          name: sketcher_${{ matrix.config.name }}
-          path: ${{ matrix.config.app }}
+          name: sketcher_${{ matrix.build-output }}
+          path: build/sketcher_app
```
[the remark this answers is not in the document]
ricrogz: As we talked, this was about having the same (hardcoded) paths for all output types. I'd rather prefer we have different directories depending on what we're building.

external/CMakeLists.txt
cdvonbargen: Done.

.github/workflows/sketcher-builder.yml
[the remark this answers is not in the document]
cdvonbargen: Talked offline and I asked to handle this in a future review when I incorporate the WASM build

ricrogz (approved):

PR #13, opened 2025-02-06
SKETCH-2182: Fixing Ubuntu builds

KevKeating:
* Linked Case: SKETCH-2182
* Branch: main
 
### Description
This diff gets building working on the Ubuntu runner.  The required changes are:
- Upgrading Qt to 6.7.  With Qt 6.5.3, we run into [QTBUG-119469](https://bugreports.qt.io/browse/QTBUG-119469) on both Linux and Windows.  The fix has been backported to Qt 6.5.7, but that version counts as a long-term support version, so it's only available to paid customers.  We're a paid customer so we can access it, but we should use freely available versions if we want to make the repo public.  I don't think we'll run into much in the way of differences between Qt 6.5 and Qt 6.7, so it shouldn't be difficult to keep Sketcher building with the two different versions (since we'll need to build with Qt 6.5 for use with the core suite).
- Installing a bunch of system packages.  The list of packages was taken from https://doc.qt.io/qt-6/linux-requirements.html.  These packages are required even if we're using a cached build of Qt; otherwise, the "Configure CMake" steps fails with errors about Qt being broken.
- Using `xvfb-run` to launch the tests so that we can instantiate a QApplication.
- Explicitly telling Qt that it needs to build xcb and a few xcb-related components.
- Disabling LZMA compression, since we're not providing the LZMA package to the build.  I think this would only matter if we wanted the Sketcher to be able to decompress files that were compressed using LZMA.

There's one additional change that's not strictly required, but it makes life easier: adding
```
    -DCMAKE_POLICY_DEFAULT_CMP0177:STRING=NEW
    -DCMAKE_POLICY_DEFAULT_CMP0174:STRING=NEW
```
to the `CMAKE_ARGS` for building Qt.  Without that, the build output for Qt contains several hundred lines complaining that the policies aren't set.  According to the CMake documentation, "policies" are kind of like long-term feature flags for any CMake changes that break backwards compatibility.  If you don't set the policy, you get the backwards-compatible but probably broken behavior (since they probably wouldn't have bothered to change it unless there was an issue) plus a warning every single time your build does anything that could potentially be affected by the behavior.

### Testing Done
The two automated tests now pass.


cdvonbargen (approved):
Yes!!

ethan-schrodinger (approved):

PR #14, opened 2025-02-12
SKETCH-2183: Fixing Windows builds

KevKeating:
* Linked Case: SKETCH-2183
* Branch: main
 
### Description
There were a few issues that needed fixing on Windows:

- On Windows, the Qt build must be run from `cmd` instead of `bash`.  The Qt build process has it's own time-stamping mechanism (that I assume is intended for speeding up incremental builds) that it runs by shelling out a process.  On Windows, Qt always shells out to `cmd` regardless of what shell it's running under, and the build process doesn't know enough to protect the path's backslashes from being interpreted as escape characters by `bash` when it launches the new shell.
- The RDKit build needed a few extra tweaks to convince it that we really, really, actually want a static build, not a partially static, partially shared frankenbuild.  The `-static` flag breaks the Linux and Mac builds because then it tries to statically link the shared libraries that it pulls from the OS (libpng happened to be the first one to cause an error), so I set that flag to only be applied on Windows.
- When GitHub pulls out the built executable, we need to tell it to look for a filename with a ".exe" extension on Windows.
- I also added another explicit policy flag so that the build logs wouldn't contain a warning about it not being set.
 
### Testing Done
The GitHub runners now produce artifacts for Mac, Linux, and Windows, and both unit tests pass on all three platforms.


KevKeating:
(I just figured out how to completely flush the cache from the command line.  Apparently it works better if you try to run the commands on the correct repo instead of the wrong one.)

ethan-schrodinger (approved):

KevKeating:
The tests should now pass for the PR.  There were two issues:
- The previous test run was using the cached dependency builds from the `main` branch, which didn't have these changes applied.  I completely flushed the cache, which solves the issue.
- RDKit doesn't like it when RDK_INSTALL_STATIC_LIBS is set to ON on Linux or Mac, so that value is now set to ON only on Windows.

cdvonbargen (approved):
Huzzah!

PR #15, opened 2025-03-11
SKETCH-2184: Enabling WASM builds

KevKeating:
* Linked Case: SKETCH-2184
* Branch: main
* Primary Reviewer: @username
 
### Description
There wound up being a bunch of changes necessary to get the WASM builds working:

- I added Emscripten to the "GitHub path," which is just the `$PATH` that gets set for the rest of the steps in the GitHub runner build.  (Changes to `$PATH` get thrown out otherwise since each step of the build runs in a fresh shell process.)  I also added calls to Emscripten's CMake wrapper (`emcmake`) for configuration and sourced `emsdk_env.sh` before building.
- As part of the Qt build process, either for Qt itself or for the Sketcher, we implicitly run Qt command line tools like `rcc` and `uic`.  This means that when we're building the WASM version of Sketcher, the GitHub runner needs a WASM build of Qt _and_ a standard Linux build of Qt (since Ubuntu can't run the WASM versions of the tools).  The WASM runner could use cached Qt version from the Linux runner, but there's no guarantee that it would be in the cache.  To deal with this, the WASM runner now builds its own slightly-stripped-down version of Qt for Linux.  (It includes qtbase but omits qtsvg, which isn't needed here.)  I stuck the build in `external/host` to distinguish it from the WASM build in `external`.
- The WASM build needs to be zipped up into a tarball, so I created a shell script for that and added it as a new step for the GitHub runner.  The logic is all taken from [build_shetcher_wasm.sh](https://github.com/schrodinger/sketcher-wasm/blob/9ed074db615080858ff9a3a32a267a229837f30c/src/build_sketcher_wasm.sh#L78) in the `sketcher-wasm` repo.  I've mostly just updated some of the paths to match the directory structure here and added a few `echo`s so that there's something in the log if we ever need a post-mortem.
- I copied the WASM HTML file, icons, etc from the `sketcher-wasm` repo to this one.
- Emscripten's CMake wrapper (`emcmake`) tells CMake that it's only allowed to look in certain directories for external libraries since it wants to prevent CMake from finding system libraries (built for Linux, not WASM) and trying to link against those.  As a result, I had to make a few tweaks so that CMake can find the dependencies that we build.  As a result, the `build-external` step now outputs absolute paths and there are a few spots where I add `CMAKE_CURRENT_BINARY_DIR` to `CMAKE_FIND_ROOT_PATH` (which is the list of directories that CMake is allowed to look in for any find_whatever calls).
- The `emcmake` script itself is just a simple shell script that passes `-DCMAKE_TOOLCHAIN_FILE=...` and `-DCMAKE_CROSSCOMPILING_EMULATOR=...` command line options to `cmake`.  (The "real" emscripten configuration stuff is located in the two files specified by those options.)  However, external projects are run in an entirely separate CMake session, and those settings don't get inherited since external projects intentionally don't inherit _any_ settings.  To fix that, I added some logic so that those two options get explicitly passed to all of the external projects if they're set.  Otherwise, all of the external projects wind up as Linux builds instead of WASM builds.
- Building `boost::beast` depends on a bunch of other `boost` libraries, including `boost::context`, which uses some i386 assembly code that won't build under Emscripten.  (Although I admittedly didn't spend too long trying to get it to build.)  To get around this, I've set up `boost::beast` to "build" (i.e. copy) as a headers-only library, which lets us use it without pulling in the other libraries.  I also set `BOOST_BEAST_HEADER_ONLY` during the build, which causes `boost::beast` to `#include` its implementation files in the headers.
- The `-DINPUT_opengl=no` flag causes Emscripten builds of Qt to fail, so that flag is now only applied to non-WASM builds.  It looks like the WASM versions of Qt winds up using OpenGL for some things (e.g. rendering) since the usual desktop interfaces aren't available, so I'm assuming that's why the flag causes issues.
- I added a few other compilation flags for Emscripten builds, mostly because compiler error messages told me to.
- We previously had a patch that prevents the zlib build from trying to build the shared version of the libraries, but I had to modify it to prevent the shared build from even being defined.  Otherwise, the Emscripten zlib build fails when the static build target and shared build taget both try to define rules for the same file (libz.a), since Emscripten doesn't define a separate file extension for shared libraries (since it doesn't have shared libraries).
- RDKit's InChi build tries to guess whether the C compiler provides `isascii()` (which is the standard) or `__isascii()` (which appears to have been a thing in old Unixes, I think?), but it guesses wrong for Emscripten builds.  To deal with this, I've patched our RDKit build to pass `-D__isascii=isascii` when building an Emscripten version of InChi, which means that the wrong guess will still wind up with the right function name.
- I got rid of the `UPDATE_DISCONNECTED 1` lines in external\CMakeLists.txt.  Those lines prevent `PATCH_COMMAND` from being run if Emscripten thinks that the patch has already been applied.  Instead, I've just prefaced `PATCH_COMMAND`s with `git reset --hard HEAD &&`, which means that we'll always remove any existing patches and then apply (or re-apply) the current ones.  This shouldn't cause any meaningful delays in the build process, both because the patch commands run almost instantaneously and because we'll typically pull the packages out of the cache without needing to build/patch/etc.  Without this change, modifying the contents of a patch winds up being a multi-step process and it's easy to (repeatedly) forget a step.

I haven't tried running the unit tests for the Emscripten build, although I think it might be possible to do so using Node.js.  ~~Should I file a separate case for getting that working and/or look into it as part of this case?~~ Update: I filed SKETCH-2388 and posted a PR (#16).

### Testing Done
All four builds now run correctly.


ethan-schrodinger (approved):

.github/actions/activate-emsdk/action.yml
```diff
@@ -21,4 +21,6 @@ runs:
       run: |
         cd build/external/emsdk-${{ inputs.version }}/emsdk/ && \
         ./emsdk install ${{ inputs.version }} && \
-        ./emsdk activate ${{ inputs.version }}
\ No newline at end of file
+        ./emsdk activate ${{ inputs.version }} && \
+        echo "${{github.workspace}}/build/external/emsdk-${{ inputs.version }}/emsdk/" >> $GITHUB_PATH && \
+        echo "${{github.workspace}}/build/external/emsdk-${{ inputs.version }}/emsdk/upstream/emscripten" >> $GITHUB_PATH
```
cdvonbargen, line 26: Is it reasonable to pull this out into a separate step? `- name: Add emscripten to GitHub PATH` or something like that?

.github/actions/build-external/action.yml
```diff
@@ -8,27 +8,41 @@ inputs:
...
+    required: false
+    default: false
 
 outputs:
   path:
-    description: "Path to the installed dependency"
+    description: "Absolute path to the installed dependency"
     # FIXME: collapse this with cache-check.path
-    value: build/external/${{ inputs.library }}-${{ inputs.version }}
+    value: ${{ github.workspace }}/build/external/${{ inputs.host-build == 'true' && 'host/' || '' }}${{ inputs.library }}-${{ inputs.version }}
```
cdvonbargen, line 20: I wonder if `build/external/${{ inputs.host-build == 'true' && 'host/' || '' }}` can get factored out in some way.

I don't know how to manipulate the action yml well enough to make a useful suggestion, so I'm fine with leaving this based on my own ignorance.

.github/workflows/sketcher-builder.yml
```diff
@@ -55,53 +55,54 @@ jobs:
           version: ${{ fromJson(env.VERSIONS_JSON).emscripten }}
 
       - name: Configure Dependency CMake
-        if: matrix.build-output != 'wasm'
         run: |
-            cmake -B build/external/ -S external/ -G Ninja \
+            ${{ matrix.build-output == 'wasm' && 'emcmake ' || '' }}\
```
cdvonbargen, line 59: same question here, can `${{ matrix.build-output == 'wasm' && 'emcmake ' || '' }}` be factored out into a variable

external/rdkit_inchi_isascii.patch
```diff
@@ -0,0 +1,17 @@
+diff --git a/External/INCHI-API/CMakeLists.txt b/External/INCHI-API/CMakeLists.txt
```
cdvonbargen, line 1: I asked this before, but now I don't remember -- should we be patching RDKit directly with this?

external/zlib_no_shared.patch
```diff
@@ -0,0 +1,68 @@
+diff --git a/CMakeLists.txt b/CMakeLists.txt
+index 0fe939d..b14cc88 100644
+--- a/CMakeLists.txt
++++ b/CMakeLists.txt
```
cdvonbargen, line 4: sigh, this is annoying, thank you for figuring it out

cdvonbargen (approved):

.github/actions/build-external/action.yml
KevKeating: Yeah, I'm not a huge fan of the way this is written either, since that same value is also set separately in both of the steps below.  I tried to figure out some way to clean this up or at least remove the duplication, but I couldn't get anything to work.  My initial thought was to create an `env` block that would apply to the whole file, but the contents of `env` blocks appear to only get evaluated once, not once per call into the file, so anything that needs to vary per-build winds up wrong.  The GitHub actions documentation wasn't any help here, so I eventually gave up.

PR #16, opened 2025-03-12
SKETCH-2388: Enabling tests on WASM builds

KevKeating:
* Linked Case: SKETCH-2388
 
### Description
This diff runs the C++ unit tests for the WASM build by using Node.js.  I stumbled across some stuff about this while I was trying to get the WASM build working in the first place, and it turns out that the changes required were very minimal: basically just telling GitHub to set up Node.js and telling CMake to use it for the tests.

Do we need to get any sort of internal approval/blessing/whatever to use Node.js?  It's MIT licensed, so we should be fine there.

### Testing Done
The GitHub runners run successfully.


ethan-schrodinger (approved):

cdvonbargen (approved):
Awesome, Dan has been asking for this for years now!

> Do we need to get any sort of internal approval/blessing/whatever to use Node.js? It's MIT licensed, so we should be fine there.

I don't think so; my understanding is that you can use MIT for any purpose (including commercial), and this is now an open-source public repo.

ZontaNicola (approved):

PR #20, opened 2025-05-24
SKETCH-2426. Update RDKit to 2025.3.2

cdvonbargen:
* Linked Case: SKETCH-2426
* Branch: main
 
### Description
Updated RDKit to the version used in the suite. The static build patch needed to be updated (it no longer cleanly applied), and a couple of headers needed `#include <cstdint>` to compile properly on windows-latest

Also added Eigen as a test dependency (at least one of the rdkit_extensions tests uses it)

When looking at building all of rdkit_extensions, found a weird include error where `#include <rdkit/GraphMol/GraphMol.h>` needs `${RDKit_INCLUDE_DIRS}` to be directly on the include path as well 🫤 

### Testing Done
CI Runners


external/rdkit_cstdint.patch
```diff
@@ -0,0 +1,24 @@
+diff --git a/Code/ForceField/MMFF/StretchBend.h b/Code/ForceField/MMFF/StretchBend.h
+index 4eceb3e7e..765fabde6 100644
+--- a/Code/ForceField/MMFF/StretchBend.h
++++ b/Code/ForceField/MMFF/StretchBend.h
+@@ -10,6 +10,7 @@
+ #ifndef __RD_MMFFSTRETCHBEND_H__
+ #define __RD_MMFFSTRETCHBEND_H__
+ 
++#include <cstdint>
```
KevKeating, line 9: Should we push these changes upstream (and keep using this patch for now)?  I got similar errors when I was testing local builds of RDKit.  I assumed it was something weird about my setup, but it seems like it's a real problem if the GitHub builder is having issues.

KevKeating (approved):

external/rdkit_cstdint.patch
cdvonbargen: Yeah, will do.

ethan-schrodinger (approved):

PR #21, opened 2025-05-29
SKETCH-2186: Adding memtests and a valgrind runner

KevKeating:
* Linked Case: SKETCH-2186
* Branch: main
 
### Description
With this diff, the CMake file will create valgrind tests whenever the build type is set to Debug.  The diff also adds a new GitHub workflow job that builds a Debug build and runs all of the valgrind tests, assuming that the regular tests pass.  I've added two suppression files: one that's copied from the Qt suppressions in mmshare ([this file](https://opengrok.schrodinger.com/xref/mmshare/build_tools/valgrind_Qt_suppressions.Linux-x86_64?r=50b6fbf1)) and one that contains the suppressions generated by the GitHub runner.  The GitHub runner is using a different version of Qt and different XCB libraries, etc, so the suppressions aren't the same.  The file from mmshare will be relevant when the valgrind tests are run locally, though.

### Testing Done
Valgrind tests pass on the GitHub runner.

### DIFF
--- a/.github/workflows/sketcher-builder.yml
+++ b/.github/workflows/sketcher-builder.yml
@@ -7,9 +7,6 @@ on:
   schedule:
     - cron: "0 4 * * *"
 
-env:
-  BUILD_TYPE: Release
-
 jobs:
   build:
     runs-on: ${{ matrix.os }}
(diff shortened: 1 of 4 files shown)


ethan-schrodinger (approved):
Don't know anything about this stuff

Seems okay to me

PR #23, opened 2025-05-30
Adding a unit test to compare dependency versions between public builds and core suite builds

KevKeating:
* Linked Case: SKETCH-2429
* Branch: main
 
### Description
This diff adds a Python unit test that compares the version numbers listed in `external/versions.json` to those in `mmshare`'s package factory.  The test can't be launched through the `premake` file (or the Makefile that generates), since `buildinger` strips `SCHRODINGER_PACKAGE_FACTORY_BASE` from the environment, but it can be launched from the command line via `$SCHRODINGER/utilities/py.test test/test_dependency_versions.py`.  Presumably, once we have an automated builder to build the core suite version of Sketcher, we can have that same machine/runner/whatever run the unit test as well.

### Testing Done
Ran the test with a recent pull of `master` on `mmshare`.

### DIFF
--- a/test/test_dependency_versions.py
+++ b/test/test_dependency_versions.py
@@ -1,0 +1,142 @@
+"""
+A unit test to ensure that the versions of dependency packages stay in sync
+between open-source builds of Sketcher and the builds used in the Schrodinger
+core suite.
+
+Note that this test should **not** be launched from premake or the generated
+Makefile. Buildinger strips $SCHRODINGER_PACKAGE_FACTORY_BASE from the
+environment when it executes those, which means that the unit test may not be
+able to find the pixi executable, which it needs to fetch version information.
+The test should instead be launched directly from a terminal via::
+
+    $SCHRODINGER/utilities/py.test test_dependency_versions.py
+"""
+
+import itertools
+import json
+import logging
+import os
+import pathlib
+import re
+import sys
+import types
+from collections.abc import Iterable
+
+SKETCHER_TEST_DIR = pathlib.Path(__file__).parent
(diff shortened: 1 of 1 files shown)


test/test_dependency_versions.py
```diff
@@ -0,0 +1,142 @@
...
+    version_text = version_text.removeprefix("release")
+    version_text = version_text.lstrip("_")
+    # ignore any fields that contains letters, as well as anything afterwards
+    version_fields = re.split(r"[._]", version_text)
+    numeric_fields = itertools.takewhile(lambda field: field.isdigit(),
+                                         version_fields)
+    return tuple(map(int, numeric_fields))
+
+
+def test_dependency_versions():
```
cdvonbargen, line 124: maybe this is silly -- could you load sketcher_versions into a global and parameterize the test on those versions, and then xfail the packages in PACKAGES_TO_IGNORE? (Maybe not empscripten, but I kinda want a report when running this that Qt, boost, and fmt are out of sync)

test/test_dependency_versions.py
```diff
@@ -0,0 +1,142 @@
...
+# Packages that we should exclude from the version checks
+PACKAGES_TO_IGNORE = [
+    # Core suite builds frequently use a commercial-only LTS release of Qt,
+    # which are not available for open-source builds of Sketcher
+    "qt",
+    # The core suite doesn't include emscripten
+    "emscripten",
+    # this repo is currently on a newer RDKit and Boost versions than the core
+    # suite
+    "rdkit",
```
cdvonbargen, line 39: oh! I updated rdkit in https://schrodinger.atlassian.net/browse/SKETCH-2426 -- this is one of the ones I want a report that it's out of sync so I remember to update when Ricardo updates in the suite

PR #24, opened 2025-05-31
SQLite::SQLite3, RDKit::MarvinParser, Boost::iostreams

cdvonbargen:
* Linked Case: SKETCH-2424
* Branch: 2025-3
 
### Description
Last few updates to get rdkit_extensions ready to be moved into the public repo. This adds building and linking sqlite, linking MarvinParse, and making sure that iostreams is linked properly. I added example calls for all 3 of these to confirm they are working. Annoyingly, boost/iostream's CMake files don't work for static building in 1.83.0 (and are fixed in 1.87.0) so I included a patch until we update boost.

A few other misc changes:
- I missed added ${COMMON_EXTERNAL_PROJECT_CMAKE_ARGS} to Eigen in the last review
- I updated CMake to 3.30 which CMP0167 requires it, and fixed the cmake version the runners use
- Updated the formatting check to run on only git files (so it doesn't run on files in build/external/)

### Testing Done
CI


KevKeating (approved):

CMakeLists.txt
```diff
@@ -1,4 +1,4 @@
-cmake_minimum_required(VERSION 3.24)
+cmake_minimum_required(VERSION 3.30)
```
KevKeating, line 2: You didn't merge in the switch back to 3.24 here (and in `external/CMakeLists.txt`.  That's why the debug builder is failing.

.github/workflows/sketcher-builder.yml
```diff
@@ -34,10 +34,10 @@ jobs:
           echo PATH=$PATH >> $GITHUB_ENV
 
       - name: Install Python packages
-        run: python3 -m pip install ninja clang-format==14.0.6
+        run: python3 -m pip install cmake==3.30.0 ninja clang-format==14.0.6
```
KevKeating, line 37: The runners already have CMake 3.31 on them, so you don't need this (and I'm not sure if it's contributing to the debug build failure).  I'm surprised that merging my branch in didn't revert this change.

cdvonbargen: Done.

PR #31, opened 2025-08-01
SKETCH-2496. Move nightly builds out to new workflow with version bump

cdvonbargen:
* Linked Case: SKETCH-2496
* Branch: 2025-4
 
### Description
This creates a new workflow that runs a nightly build for the 2025-4 release which:
- if there were commits in the last 24 hours, bumps the patch value of VERSION in CMakeLists.txt
- runs the sketcher-builder.yml workflow
The idea here is we'll need to create another 2026-1.yml workflow when we branch for 25-4, as well as a need to capture a build number in some way like we do with version.h -- I'm open to other ideas (I considered pulling the GHA build number and stuffing that into the patch version? but that seemed a little goofy?)

### Testing Done
Tested locally in my repo fork

### DIFF
--- a/.github/workflows/2025-4.yml
+++ b/.github/workflows/2025-4.yml
@@ -1,0 +1,71 @@
+name: 2025-4 Nightly Build
+
+on:
+  schedule:
+    - cron: '0 2 * * *'
+  workflow_dispatch: ~
+
+jobs:
+  increment-version:
+    runs-on: ubuntu-latest
+    outputs:
+      has_recent_commits: ${{ steps.check-commits.outputs.has_recent_commits }}
+    steps:
+      - uses: actions/checkout@v4
+
+      - name: Configure Git
+        run: |
+          git config --local user.email "action@github.com"
+          git config --local user.name "GitHub Action"
+
+      - name: Check for recent commits
+        id: check-commits
+        run: |
+          # Check if there are commits in the last 24 hours
+          recent_commits=$(git log --since="24 hours ago" --oneline | wc -l)
+          echo "Recent commits in last 24 hours: $recent_commits"
+
+          if [ $recent_commits -gt 0 ]; then
+            echo "has_recent_commits=true" >> $GITHUB_OUTPUT
+            echo "Found $recent_commits commits in the last 24 hours, proceeding with version increment"
+          else
+            echo "has_recent_commits=false" >> $GITHUB_OUTPUT
+            echo "No commits in the last 24 hours, skipping version increment"
+          fi
+
+      - name: Increment patch version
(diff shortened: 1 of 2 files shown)


.github/workflows/2025-4.yml
```diff
@@ -0,0 +1,71 @@
...
+          grep "VERSION " CMakeLists.txt
+
+          echo "NEW_VERSION=$new_version" >> $GITHUB_ENV
+
+      - name: Commit version change
+        if: steps.check-commits.outputs.has_recent_commits == 'true'
+        run: |
+          git add CMakeLists.txt
+          git commit -m "Increment patch version to ${{ env.NEW_VERSION }}"
+          git push
```
lennax, line 66: is the default branch protected? if so, there might need to be special logic to allow this automation to bypass the branch protections.

.github/workflows/2025-4.yml
```diff
@@ -0,0 +1,71 @@
...
+      - name: Configure Git
+        run: |
+          git config --local user.email "action@github.com"
+          git config --local user.name "GitHub Action"
+
+      - name: Check for recent commits
+        id: check-commits
+        run: |
+          # Check if there are commits in the last 24 hours
+          recent_commits=$(git log --since="24 hours ago" --no-merges --oneline | wc -l)
```
lennax, line 25: does this need logic to avoid the previous day's "increment patch" commit?

.github/workflows/2025-4.yml
```diff
@@ -0,0 +1,71 @@
+name: 2025-4 Nightly Build
+
+on:
+  schedule:
+    - cron: '0 2 * * *'
```
lennax, line 5: github recommends avoiding scheduling workflows for minute 0 to avoid demand peaks

https://docs.github.com/en/actions/reference/workflows-and-actions/events-that-trigger-workflows#schedule

KevKeating (commented):
Looks good to me outside of the issues that Lenna pointed out.

.github/workflows/2025-4.yml
cdvonbargen: Done.

.github/workflows/2025-4.yml
cdvonbargen: yeah, it specifically looks for `version_commits=$(git log --since="24 hours ago" --grep="Increment patch version" --oneline | wc -l)` which get removed from the list of commits it considers

cdvonbargen:
When we made the repo private for a hot minute this branch got disconnected and the PR closed. Opening a new PR here: https://github.com/schrodinger/sketcher/pull/72

PR #37, opened 2025-08-14
rework selection tools

ZontaNicola:
rework selection tools and add ellipse

* Linked Case: SKETCH-2452
* Branch: <branch>
 
### Description
This makes the two selection tool buttons popup buttons, each allowing the selection of either square, lasso or ellipse. The default are square and lasso, like before.
I think we're missing an icon for the ellipse, but otherwise everything seems to work

### Testing Done
tested in molviewer

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -33,3 +33,4 @@ TAGS
 .pytest_cache
 # compilation database
 compile_commands.json
+.vscode/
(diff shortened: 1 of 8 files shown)


ethan-schrodinger (commented):
These changes LGTM.
>I think we're missing an icon for the ellipse, but otherwise everything seems to work

Does this mean that we're blocked by UX?

PR #39, opened 2025-08-14
SHARED-11702: Move python tests to C++

rachelnwalker:
* Linked Case: SHARED-11702
* Branch: main
* Primary Reviewer: @ricrogz 
 
### Description
This PR rewrites a majority of the python tests in test_coarsegrain_conversions.py so that we can have the testing for the atomistic <->monomeric conversions in this same repository as the implementation.

When I merge this down, I will remove the python tests that are now duplicated in C++.

### Testing Done
See newly added tests


rachelnwalker:
I was a little worried about these tests taking a significant amount of time, looks like about 3-4s normally and up to a minute for the memtest.

Looking into the memtest failures that looks unrelated

ricrogz (approved):
I added some minor suggestions, but LGTM.

PR #50, opened 2025-08-28
SKETCH-2091 add option to copy as image

ZontaNicola:
* Linked Case: SKETCH-2091
* Branch: 2026-1
 
### Description
This adds the option to copy images and reactions as an image. It is active from both the selection context menu and the general background context menu (Copy All As...)

Full Reactions are rendered if a reaction is present and
-we're using the copy all as 
-we are exporting a selection, and the selection includes at least one non molecular object.

If the user selects only molecules on a reaction, we export just the selection as a molecule (the idea is that someone wants to copy a product or reactant), but if the selection includes non molecular objects we assume they mean the whole reaction (since we can't export parts of the reaction) and give them that

### Testing Done
tested by copying as image and saving into a google doc

### DIFF
--- a/src/schrodinger/sketcher/menu/cut_copy_action_manager.cpp
+++ b/src/schrodinger/sketcher/menu/cut_copy_action_manager.cpp
@@ -77,20 +77,18 @@ void CutCopyActionManager::initCopyAsMenu()
             // set a flag on the action to determine its visibility later
             action->setData(QVariant(is_reaction_format));
         }
-        // Add a separator and the option to export as an image
-        auto separator = m_copy_as_menu->addSeparator();
-        separator->setData(QVariant(is_reaction_format));
-        auto action = m_copy_as_menu->addAction(
-            "Image", this, [this]() { emit copyAsImageRequested(); });
-        // export as image should only be allowed for full scene
-        m_hide_for_selections.push_back(separator);
-        m_hide_for_selections.push_back(action);
-        // set a flag on the action to determine its visibility later
-        action->setData(QVariant(is_reaction_format));
     };
 
     init_menu(get_standard_export_formats(), false);
     init_menu(get_reaction_export_formats(), true);
+
+    // Add a separator and the option to export as an image
+    auto separator = m_copy_as_menu->addSeparator();
+    auto action = m_copy_as_menu->addAction(
+        "Image", this, [this]() { emit copyAsImageRequested(); });
+    // export as image should only be allowed for full scene
+    m_hide_for_selections.push_back(separator);
+    m_hide_for_selections.push_back(action);
 }
 
 void CutCopyActionManager::updateActions()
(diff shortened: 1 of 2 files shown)


src/schrodinger/sketcher/resources/fonts/Arial.ttf
KevKeating, line 1: You should remove these fonts.

src/schrodinger/sketcher/menu/cut_copy_action_manager.cpp
```diff
@@ -76,6 +76,13 @@ void CutCopyActionManager::initCopyAsMenu()
...
             action->setData(QVariant(is_reaction_format));
         }
+        // Add a separator and the option to export as an image
+        auto separator = m_copy_as_menu->addSeparator();
+        separator->setData(QVariant(is_reaction_format));
+        auto action = m_copy_as_menu->addAction("Image", this, [this]() {
+            emit copyAsImageRequested(getSubset());
+        });
+        action->setData(QVariant(is_reaction_format));
     };
```
KevKeating, line 87: This is only tangentially related to your change, but it might be helpful to have a comment here mentioning that the action's data is used to control visibility in the updateActions method.  At first, I was confused as to why you were setting data on the separator.

src/schrodinger/sketcher/sketcher_widget.cpp
```diff
@@ -374,6 +374,25 @@ void SketcherWidget::copy(Format format, SceneSubset subset)
...
+    bool get_full_reaction = (m_mol_model->hasReactionArrow() &&
+                              (subset == SceneSubset::ALL ||
+                               m_mol_model->hasSelectedNonMolecularObjects()));
+    if (get_full_reaction) {
+        auto reaction = getRDKitReaction();
+        image_bytes = get_image_bytes(*reaction, ImageFormat::PNG);
+    } else {
+        auto mol = extract_mol(m_mol_model, subset);
+        image_bytes = get_image_bytes(*mol, ImageFormat::PNG);
+    }
```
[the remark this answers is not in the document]
ZontaNicola: doesn't look like that would work on selections, right?

src/schrodinger/sketcher/menu/cut_copy_action_manager.cpp
ZontaNicola: added a comment

ethan-schrodinger (commented):
>If the user selects only molecules on a reaction, we export just the selection as a molecule (the idea is that someone wants to copy a product or reactant), but if the selection includes non molecular objects we assume they mean the whole reaction (since we can't export parts of the reaction) and give them that

It probably won't ever be an issue, but I always worry about non-intuitive differences in behavior. My guess is that this peculiarity is not documented anywhere.

Specifically, though, I'm curious about this:
>we can't export parts of the reaction

Is there a reason we can export different molecules in a system, but not different parts of a reaction?

Is there a way to implement this interface that avoids this awkwardness? Would it be better to always export the entire scene?

ZontaNicola:
> The last two commit messages were "wip". Should I wait to review?

this is ready for review, as indicated by the second wip (work is perfected) comment ;)

PR #63, opened 2025-09-10
when unable to trim a bond line, keep it as is

ZontaNicola:
* Linked Case: SKETCH-2501
* Branch: <branch>
 
### Description
this prevents the maestro crash.

### Testing Done
maestro no longer crashes,
here's the resulting overlay picture:

Screenshot 2025-09-10 at 10 45 57

### DIFF
--- a/src/schrodinger/sketcher/molviewer/coord_utils.cpp
+++ b/src/schrodinger/sketcher/molviewer/coord_utils.cpp
@@ -298,13 +298,13 @@ void trim_line_to_rect(QLineF& line, const QRectF& rect)
 
     QPointF inter_point; // the intersection point
 
-    // if the line is completely inside the rect, make it an invalid line
+    // if the line is completely inside the rect, skip the trimming
     if (enlarged_rect.contains(line.p1()) &&
         enlarged_rect.contains(line.p2())) {
-        line.setP1(line.p2());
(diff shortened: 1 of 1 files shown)


KevKeating:
> I'd rather have an ugly picture than a wrong picture

Yeah, that's definitely a valid point.

> one idea would be to have a minimum resulting size when shortening a bond

I really like this idea.  It seems like it would work well both from the "make the least ugly picture possible" perspective as well as the "wouldn't be too difficult to implement" perspective.

src/schrodinger/sketcher/molviewer/coord_utils.cpp
```diff
@@ -305,13 +305,24 @@ void trim_line_to_rect(QLineF& line, const QRectF& rect, qreal min_length)
     }
     if (enlarged_rect.contains(line.p1()) &&
         intersection_of_line_and_rect(line, enlarged_rect, inter_point)) {
-        if (QLineF(inter_point, line.p2()).length() >= min_length) {
+        if (QLineF(inter_point, line.p2()).length() < min_length) {
+            // trim to exactly min_length. We use a temporary line in the
+            // opposite direction because setLenght() moves p2 and we want to
```
KevKeating, line 310: setLenght -> setLength

KevKeating (approved):

PR #65, opened 2025-09-11
SKETCH-2532: Prevent the settings dialog from overwriting the color scheme

KevKeating:
* Linked Case: SKETCH-2532
* Branch: main
 
### Description
The color scheme was being overwritten when:
  - the `SketcherWidget` was instantiated
  - the color ccheme was immediately (before the event loop had a chance to run) changed via `SketcherWidget::setColorScheme`
  - control flow returned to the event loop (or the event loop was started)

The cause of this was the Preferences dialog, oddly enough.  Instantiating the dialog would load the default settings into the dialog's widgets.  Because the widgets changed, the update timer would start.  That timer would fire as soon as control flow went back to the event loop, at which point the dialog would write these default settings back to the `SketcherModel`.  This effectively meant that the dialog would overwrite any (dialog-relevant) settings that had been changed in between the `SketcherWidget` instantiation and control flow returning to the event loop.  I've now added a line to the dialog's constructor that stops the timer, which solves the issue.  Since the settings came from the `SketcherModel` in the first place, there's no actual need to write them back to the model.

### Testing Done
I instantiated a `SketcherWidget()` in an ipython terminal and confirmed that DARK_MODE survives the event loop starting.  I also added a new unit test that only passes with the `rendering_setting_dialog` changes.


src/schrodinger/sketcher/molviewer/abstract_atom_or_bond_display_settings.cpp
```diff
@@ -0,0 +1,85 @@
...
+}
+
+RDKit::DrawColour get_rdkit_dark_mode_carbon_color()
+{
+    RDKit::ColourPalette palette;
+    RDKit::assignDarkModePalette(palette);
+    auto carbon = static_cast<int>(Element::C);
+    carbon = palette.contains(carbon) ? carbon : -1;
+    return palette[carbon];
+}
```
KevKeating, line 82: None of the logic here is actually new; it's just been moved from atom_display_settings.cpp.

KevKeating:
Matvey noticed that `setColorScheme` wasn't updating the bond color.  I hadn't run into that because activating a ring tool fixes the issue (for reasons that I'm still not entirely sure of), and I'd been using the ring tool to draw a quick test structure.  The cause of the issue was that the logic for setting the bond color (which is just a one-liner) was in the Properties dialog code instead of the SketcherModel and/or the display settings classes.  I've now done a bit of refactoring to the display settings classes so that BondDisplaySettings knows how to update its bond color based on a color scheme, which avoids the issue and makes sure that settings logic lives in a model class instead of a dialog.  As part of that, I created a new AbstractAtomOrBondDisplaySettings class to hold the logic that's now shared between AtomDisplaySettings and BondDisplaySettings.

KevKeating:
I just pushed a minor tweak: the `m_update_timer->stop()` call is now part of `RenderingSettingsDialog::loadSettingsFromModel` instead of being called from the `RenderingSettingsDialog` constructor immediately _after_ the call to `loadSettingsFromModel`.  This prevents the same issue from happening after other calls to `loadSettingsFromModel` (which seems unlikely to occur in practice, but would be theoretically possible).

cdvonbargen (approved):

PR #66, opened 2025-09-13
Restore builds; pin cmake; skip rdkit_cstdint.patch

cdvonbargen:
* Branch: 2025-4
* Primary Reviewer: @ricrogz 
 
### Description
I cleared the dep caches, and the mac and ubuntu build fail. The mac build fails because it looks like the cmake version on macos-latest was *too* new that the zstd build was failing. And @ricrogz 's patch for extern created weird errors on ubuntu. This change does the following:
- pin cmake (3.24) and ninja (1.13.0) versions
- comment out CMP0167 policy (it only exists in 3.30!)
- skip rdkit_extern_patch.patch (for now)
I'm working on fixing the extern thing (Brian left comments on https://github.com/rdkit/rdkit/pull/8765) but these changes allow the builds to complete and the non-wasm tests to pass. Second review to come.

Also a couple of other things:
- added boost::locale in anticipation of updating RDKit
- Remove PARENT_SCOPE -- it was creating warnings in the configure stage
- Ran cmake-format on CMakeLists.txt files

### Testing Done
CI

### DIFF
--- a/.github/workflows/sketcher-builder.yml
+++ b/.github/workflows/sketcher-builder.yml
@@ -34,7 +34,7 @@ jobs:
           echo PATH=$PATH >> $GITHUB_ENV
 
       - name: Install Python packages
-        run: python3 -m pip install ninja clang-format==14.0.6
+        run: python3 -m pip install clang-format==14.0.6 cmake==3.24 ninja==1.13.0
 
       - name: Install valgrind
         if: matrix.build-output == 'memtest'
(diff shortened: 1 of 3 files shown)


ricrogz (commented):
LGTM except for the disabling of the externs patch.

cdvonbargen:
Blarg -- I accidentally deleted the branch -- recreated the review with the review comments here: https://github.com/schrodinger/sketcher/pull/69

PR #67, opened 2025-09-15
SKETCH-2453 clenup/fit to works on selection

ZontaNicola:
* Linked Case: SKETCH-2453
* Branch: <branch>
 
### Description
the two toolbar actions now work on selections when they are present (and have the correct green bg)

### Testing Done
tested in molviewer.

### DIFF
--- a/src/schrodinger/sketcher/molviewer/view.h
+++ b/src/schrodinger/sketcher/molviewer/view.h
@@ -36,10 +36,16 @@ class SKETCHER_API View : public QGraphicsView
 
   public slots:
     /**
-     * Scale and center the view matrix so that all objects in the scene are
+     * Scale and center the view matrix so that objects in the scene are
      * visible, but avoid zooming in too much for small molecules
+     * @param selection_only If true, only consider selected objects
      */
-    void fitToScreen();
+    void fitToScreen(bool selection_only = false);
+
+    void fitAllToScreen()
+    {
+        fitToScreen(false);
+    }
 
     /**
      * If needed scale the view matrix to include objects that are off-screen.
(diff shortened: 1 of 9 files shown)


src/schrodinger/sketcher/molviewer/view.h
```diff
@@ -36,10 +36,16 @@ class SKETCHER_API View : public QGraphicsView
...
      * visible, but avoid zooming in too much for small molecules
+     * @param selection_only If true, only consider selected objects
      */
-    void fitToScreen();
+    void fitToScreen(bool selection_only = false);
+
+    void fitAllToScreen()
+    {
+        fitToScreen(false);
+    }
```
KevKeating, line 48: Do we need this method if fitToScreen's default value for `selection_only` is `false`?  You should be able to connect `MolModel::newMoleculeAdded` directly to `fitToScreen`.

src/schrodinger/sketcher/sketcher_widget.cpp
```diff
@@ -493,10 +493,11 @@ void SketcherWidget::connectTopBarSlots()
             &QUndoStack::undo);
     connect(m_ui->top_bar_wdg, &SketcherTopBar::redoRequested, m_undo_stack,
             &QUndoStack::redo);
-    connect(m_ui->top_bar_wdg, &SketcherTopBar::cleanupRequested, [this]() {
-        m_mol_model->regenerateCoordinates();
-        m_ui->view->fitToScreen();
-    });
+    connect(m_ui->top_bar_wdg, &SketcherTopBar::cleanupRequested,
+            [this](bool selection_only) {
+                m_mol_model->regenerateCoordinates(selection_only);
```
KevKeating, line 498: I think it might be a little clearer to have something like
```cpp
if (selection_only) {
    m_mol_model->cleanUpSelection();
} else {
    m_mol_model->regenerateCoordinates();
}
```
here and get rid of the `selection_only` parameter to `regenerateCoordinates`.  It seems a little weird for `regenerateCoordinates` to have a parameter that causes it to just call a different method instead of itself.

KevKeating (commented):
I just had two minor nits.

src/schrodinger/sketcher/molviewer/view.h
ZontaNicola: yeah that's what I thought too! but apparently default values don't work that way for connect, and I had to add this function.

cdvonbargen (approved):

PR #72, opened 2025-09-16
SKETCH-2496. Move nightly builds out to new workflow with version bump

cdvonbargen:
* Linked Case: SKETCH-2496
* Branch: 2025-4
* Primary Reviewer: @lennax  
 
### Description
This creates a new workflow that runs a nightly build for the 2025-4 release which:
- if there were commits in the last 24 hours, bumps the patch value of VERSION in CMakeLists.txt
- runs the sketcher-builder.yml workflow
The idea here is we'll need to create another 2026-1.yml workflow when we branch for 25-4, as well as a need to capture a build number in some way like we do with version.h -- I'm open to other ideas though

### Testing Done
Tested locally in my repo fork


.github/workflows/2025-4.yml
juzerzarif, line 1: You shouldn't need a separate workflow file for each release line if you make the jobs in this workflow parameterized. If we make the workflow_call trigger for `sketcher-builder.yml` accept a ref as input, you can call both `increment-version` and `build-and-test` with matrix strategies with multiple branches.

.github/workflows/nightly-build.yml
```diff
@@ -1,10 +1,76 @@
...
+        required: false
+        type: string
 
 jobs:
-  placeholder:
+  nightly-build:
     runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        branch: [main, 2025-4]
```
lennax, line 21: If we're going to trigger this from the regular NB bump process, the `schedule` needs to be removed and `branch` needs to be an input

cdvonbargen: done.

PR #76, opened 2025-09-18
SKETCH-2540: Adding SketcherWidget::activateSelectionTool

KevKeating:
* Linked Case: SKETCH-2540
* Branch: main
 
### Description
This adds a SketcherWidget method for activating any of the selection tools.  The only minor caveat here is that doing something like
```py
sk = sketcher.SketcherWidget()
sk.activateSelectionTool(sketcher.SelectionTool.ELLIPSE)
sk.setSelectOnlyMode(True)
```
won't actually activate the ellipse selection tool, since "normal" Sketcher (i.e. not select only Sketcher) doesn't allow selection tools to be activated when the workspace is empty, meaning that the `activateSelectionTool()` will be a no-op.  The workaround is to make sure that select-only mode is enabled first, i.e.,
```py
sk = sketcher.SketcherWidget()
sk.setSelectOnlyMode(True)
sk.activateSelectionTool(sketcher.SelectionTool.ELLIPSE)
```
will work as expected.  This seems like a relatively minor inconvenience, especially since it's not an issue when using the (Python-only) `SelectOnlySketcherWidget` class, as that will be in select-only mode as soon as its instantiated.

I've also added SIP wrappings for the new method, but that part is mmshare-only so it's not in this diff (since we don't have the SIP wrappings in this repo yet).  It's just the new line from sketcher_widget.h copied over to _sketcher.sip.

@matvey83, GitHub still won't let me add you as a reviewer on this repo, so I'm trying a ping instead.

### Testing Done
Tested with an mmshare build using the Maestro Python terminal.

### DIFF
--- a/include/schrodinger/sketcher/sketcher_widget.h
+++ b/include/schrodinger/sketcher/sketcher_widget.h
@@ -110,13 +110,21 @@ class SKETCHER_API SketcherWidget : public QWidget
      * Enable or disable select-only mode, which removes the toolbars and limits
      * user interaction to selecting items.
      */
-    void setSelectOnlyMode(bool select_only_mode_enabled);
+    void setSelectOnlyMode(const bool select_only_mode_enabled);
 
     /**
      * Set the color scheme, which controls the atom and bond colors as well as
      * the background color.
      */
-    void setColorScheme(ColorScheme color_scheme);
+    void setColorScheme(const ColorScheme color_scheme);
+
+    /**
+     * Switch to the specified selection tool. Note that this call will have no
+     * effect if select-only mode is disabled and the workspace is empty, as the
+     * select tool isn't allowed in that scenario (since there's nothing to
+     * select).
+     */
+    void activateSelectionTool(const SelectionTool tool);
 
     /**
      * Undoably select or deselect the specified atoms and bonds.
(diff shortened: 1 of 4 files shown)


include/schrodinger/sketcher/sketcher_widget.h
```diff
@@ -110,13 +110,21 @@ class SKETCHER_API SketcherWidget : public QWidget
...
-    void setColorScheme(ColorScheme color_scheme);
+    void setColorScheme(const ColorScheme color_scheme);
+
+    /**
+     * Switch to the specified selection tool. Note that this call will have no
+     * effect if select-only mode is disabled and the workspace is empty, as the
+     * select tool isn't allowed in that scenario (since there's nothing to
+     * select).
+     */
+    void activateSelectionTool(const SelectionTool tool);
```
ethan-schrodinger, line 127: Nit: is there a reason to prefer `activateSelectionTool()` rather than `setSelectionTool()`?

ethan-schrodinger (approved):
LGTM

include/schrodinger/sketcher/sketcher_widget.h
cdvonbargen: Would it be dumb to collapse the APIs? So instead of `activateSelectionTool` and `setSelectOnlyMode` it would be:
```
void activateSelectOnlyMode(const SelectionTool tool);
```

Does anyone ever want to _unset_ select-only-mode? I can't come up with examples but if we think we need it, morph the other function into `deactivateSelectOnlyMode`.

cdvonbargen (approved):
Sweet, thank you!

PR #87, opened 2025-09-26
SKETCH-2550: Mouse cursor changes

KevKeating:
* Linked Case: SKETCH-2550
* Branch: 2025-4
 
### Description
This PR makes a handful of changes to the mouse cursor to help it feel more precise:

- The cursor outline, which provides contrast in case the cursor if over something that's the same color as the arrow, is now dramatically thinner.  That way, even if the user is looking at the tip of the outline instead of the tip of the arrow, the difference won't be substantial.  (The difference is now ~0.1 pixels, down from ~5.4 pixels, as a result of both this change and the next one.)
- The cursor outline now uses a beveled join (via SVG styling), which means that the tip of any outline corners are "chopped off."
- The cursor outline is now the exact same color as the background, instead of a light gray.
- The cursor colors are inverted when Sketcher is in dark mode.  This will hopefully make the cursor outline less noticeable, since it'll blend into the background, while the arrow itself will always contrast with the background color.
- The cursor is now scaled down by 20% for Mac desktop builds, since Mac's normally have a smaller cursor size than Windows.  I haven't actually tested whether 20% is the right value to use since I don't have a Mac.  (Although I _have_ tested the 20% scaling on Windows just to make sure that it doesn't look ridiculous.)  Would someone be able to test this and see if the cursor size inside the Sketcher window is roughly the same as the cursor size outside of it?

Here's a comparison of the old and new cursors.  Note that I've added a gray background for the new cursors, since otherwise the outline wouldn't be visible against the background.

| Old | New (light mode) | New (dark mode) |
| --- | --- | --- |
| cursor-arrow - orig | cursor-arrow | cursor-arrow - dark |

### Testing Done
Tested this on Windows using both the normal cursor size and the scaled-down-by-20%-for-Mac cursor size.


cdvonbargen (approved):
It would be good to get a build to Juzer with this fix if https://schrodinger.atlassian.net/browse/LDID8-1005 isn't done so that he can get this into a 25-4 build for Samson to test.

KevKeating:
> Why do we really need to invert the colors on dark background? The default system cursor will still be used for toolbars and also special operations like the one below, and it seems to work fine for both light and dark backgrounds (at least on Mac). Also I don't think the OS changes cursor when I set my Mac from dark to light, they simply have the black pointer with white outline that works on both. I think it might be confusing for people to see the arrow inverting colors every time they hover a bond in the example below.

This came up at standup this morning, but I figured I'd post a summary of it here for posterity's sake.  With this change, the cursor will no longer revert to the system cursor when there's no cursor hint set, so the behavior you're showing shouldn't happen.  Instead the Sketcher cursor will remain (without any hint attached) when hovering over a bond in that scenario.

My goal with inverting the colors was to make sure that the inner arrow was always highly visible and not the outline.  That way, users will (hopefully) expect the click to happen at the tip of the inner arrow, not at the tip of the outline.

I just filed https://bugreports.qt.io/browse/QTBUG-140687 to allow us to access a pixmap of the system cursor, which would mean that we could use that in place of our custom arrow cursor entirely (except on WASM builds, where I'm pretty sure we wouldn't be able to access the system cursor image due to browser security restrictions).

ZontaNicola (approved):
I still prefer to not invert colors, but it's not a big deal. Ideally we would keep the system's cursor, but no sense holding this review waiting for Qt

src/schrodinger/sketcher/model/sketcher_model.cpp
```diff
@@ -434,6 +432,12 @@ void SketcherModel::setBackgroundColor(QColor color)
     emit backgroundColorChanged(color);
 }
 
+bool SketcherModel::isDarkMode() const
+{
+    return m_color_scheme == ColorScheme::DARK_MODE ||
+           m_color_scheme == ColorScheme::WHITE_BLACK;
+}
+
```
ethan-schrodinger, line 440: Thanks for pulling this into a proper function!

Maybe we decide that this goes beyond the scope of this ticket, but it looks like [there are other places where this logic is used to determine whether "dark mode" is active](https://opengrok.schrodinger.com/xref/mmshare/src/schrodinger/sketcher/molviewer/abstract_atom_or_bond_display_settings.cpp?r=76728469#14).

ethan-schrodinger (approved):

PR #89, opened 2025-09-30
SKETCH-2543: Update public repo to use RDKit 2025.3.6 + patches

ricrogz:
* Linked Case: SKETCH-2543
* Branch: 2025-4
 
I posted the expat fix to RDKit repo as https://github.com/rdkit/rdkit/pull/8832. That part of the patch should be reverted when we start using a RDKit version that includes that PR.


ricrogz:
> It looks like the windows build is busted applying the other patch?
> 

The windows build is severely broken because of the ChemDraw external. Specificall, this file: https://github.com/Glysade/chemdraw/blob/d127473e1b334ff7020346b1f195d7d171f1ecf7/chemdraw/CoreChemistryAPI.h#L16. The `CORE_CHEMISTRY_API` macro only contemplates building for shared libs, enforcing dll import/export, which won't work for static libs.

Fixing this for the sketcher build is very complicated, because the ChemDraw sources don't exist when the patch step for the RDKit repo is executed (they are downloaded during RDKit's cmake configure step). And, of course, CMake's `ExternalProject_Add` function doesn't allow patching between the cmake configure and the build step.

So I'm just going to disable ChemDraw support for now.

external/CMakeLists.txt
```diff
@@ -234,6 +236,7 @@ ExternalProject_Add(
     -DRDK_INSTALL_STATIC_LIBS=${RDKIT_INSTALL_STATIC_LIBS_VAL}
     -DRDK_OPTIMIZE_POPCNT=OFF
     -DRDK_USE_BOOST_STACKTRACE=OFF
+    -DRDK_BUILD_CHEMDRAW_SUPPORT=OFF
```
cdvonbargen, line 239: Do we need to block .cdxml read support since this is now turned off?

ricrogz: Yes.

src/schrodinger/rdkit_extensions/convert.cpp
```diff
@@ -541,7 +541,9 @@ boost::shared_ptr<RDKit::RWMol> to_rdkit(const std::string& text,
                 // crash the Sketcher.  This #ifndef should be removed as part
                 // of SKETCH-2357.
                 Format::MRV,
-                Format::CDXML,
+        // Support for CDXML has been disabled
+        // due to build issues (SKETCH-2543)
+        // Format::CDXML,
```
rachelnwalker, line 546: Will all of these changes will also be pushed in our internal repo?

cdvonbargen: We should -- I'll ask for mgmt approval on the case

rachelnwalker: Thanks. I really don't want sketcher and internal code out of sync.

ethan-schrodinger (approved):

PR #92, opened 2025-09-30
LDID8-1005: change npm publish condition to workflow input

juzerzarif:
* Linked Case: LDID8-1005
* Branch: 2025-4
 
### Description
Two things:
- Merges the changes from [2f9f1d6](https://github.com/schrodinger/sketcher/commit/2f9f1d67e44c2c7375b9e684a5e034c4c11e2572) into `2025-4` so the release branch has the necessary source code changes to build the package
- Changes the conditional for running the npm publish step from `github.event_name == 'workflow_run'` to `inputs.publish-npm-package`. Turns out `workflow_call` is not an `event_name` because called workflows are run within the context of their caller workflow so the `github.event_name` context has the value for the event name for the caller workflow (for e.g. `workflow_dispatch` for a manually triggered nightly build workflow). So switching it over to an explicit input that the nightly build workflow provides to the sketcher-builder workflow.

### DIFF
--- a/.github/actions/publish-npm-package/action.yml
+++ b/.github/actions/publish-npm-package/action.yml
@@ -1,0 +1,35 @@
+name: publish-npm-package
+description: Publishes the @schrodinger/sketcher npm package with the WASM artifacts
+
+outputs:
+  version:
+    description: The package version that was published the npm registry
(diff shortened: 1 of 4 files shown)


cdvonbargen (approved):
Ah terrific!

PR #93, opened 2025-10-01
fix rna orientation

ZontaNicola:
* Linked Case: SKETCH-2522
* Branch: <branch>
 
### Description
I don't think we can layout proteins or nucleic acids from right to left unless it's the second chain of a double stranded nucleic acid, since we don't show the chain direction. I special cased double stranded nucleic acids so the general case doesn't invert the direction anymore.

Screenshot 2025-10-01 at 12 00 39

### Testing Done
added a test. Some coordinates changed in the current tests: here are the new pics, they look good to me but @rachelnwalker will know more
Screenshot 2025-10-01 at 11 26 29
Screenshot 2025-10-01 at 11 26 16
Screenshot 2025-10-01 at 11 26 00
Screenshot 2025-10-01 at 11 25 42

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -76,7 +76,10 @@ function(setup_target TARGET)
   endif()
   # Compilation optimization, warnings, and error handling
   if(MSVC)
-    target_compile_options(${TARGET} PRIVATE -WX)
+    target_compile_options(${TARGET} PRIVATE -WX -EHsc)
+    if(${BUILD_SHARED_LIBS})
+        target_compile_definitions(${TARGET} PRIVATE RDKIT_DYN_LINK)
+    endif()
   elseif(WIN32)
     target_compile_options(
       ${TARGET} PRIVATE -Wall -Werror -Wswitch -O2 -fvisibility=${VISIBILITY}
@@ -170,6 +173,7 @@ build_schrodinger_library(${SKETCHER_TARGET})
 target_link_libraries(
   ${SKETCHER_TARGET}
   PRIVATE Boost::boost
+          Boost::filesystem
           fmt::fmt
           Qt6::Svg
           Qt6::SvgWidgets
@@ -196,7 +200,10 @@ if(EMSCRIPTEN)
       -sDISABLE_EXCEPTION_CATCHING=0)
 endif()
 target_link_libraries(
-  ${APP_TARGET} PRIVATE Boost::boost Qt6::Widgets ${SKETCHER_TARGET}
+  ${APP_TARGET} PRIVATE Boost::boost
+                        Boost::filesystem
+                        Qt6::Widgets
+                        ${SKETCHER_TARGET}
                         ${EXTRA_SKETCHER_APP_LINK_FLAGS})
 
 # Tests
@@ -237,6 +244,10 @@ if(ENABLE_TESTING)
     target_include_directories(${TEST_TARGET}
                                PRIVATE ${PROJECT_SOURCE_DIR}/test)
     target_link_libraries(${TEST_TARGET} PRIVATE ${TEST_DEPENDENCIES})
+    if(MSVC)
+        # ignore discarded [[nodiscard]] values for the tests
+        target_compile_options(${TEST_TARGET} PRIVATE -wd4834)
+    endif()
     enable_qt_moc(${TEST_TARGET})
     if(EMSCRIPTEN)
       set(TEST_COMMAND node $<TARGET_FILE:${TEST_TARGET}>)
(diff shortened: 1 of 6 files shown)


test/schrodinger/rdkit_extensions/test_cg_coordgen.cpp
```diff
@@ -171,52 +193,54 @@ BOOST_DATA_TEST_CASE(
...
-          {1.5, -4.5},
-          {0, -4.5},
-          {0, -3},
-          {-1.5, -4.5}}},
+          {-6, -3},
+          {-6, -4.5},
+          {-4.5, -3},
+          {-3, -3},
+          {-3, -4.5},
+          {-1.5, -3}}},
```
rachelnwalker, line 202: just making sure: all of the HELM strings with differences in coordinates you included in your description, right?

src/schrodinger/rdkit_extensions/helm/cg_coordgen.cpp
```diff
@@ -694,6 +728,13 @@ unsigned int compute_monomer_mol_coords(RDKit::ROMol& monomer_mol)
     // SHARED-9795: Special case for single polymers that are strictly chains
     if (is_single_linear_polymer(monomer_mol)) {
         lay_out_snaked_linear_polymer(*polymers[0]);
+    } // SKETCH-2522: Special case for double stranded nucleic acids:
+    else if (is_double_stranded_nucleic_acid(monomer_mol, polymers)) {
+        // For double stranded nucleic acids we want to lay out the first
+        // polymer normally and then rotate the other polymer 180° so the
+        // strands run anti-parallel to each other
+        lay_out_polymer(*polymers[0]);
+        lay_out_polymer(*polymers[1], true);
```
rachelnwalker, line 737: I think making this a special case was a good call.

src/schrodinger/rdkit_extensions/helm/cg_coordgen.cpp
```diff
@@ -679,11 +686,38 @@ static bool is_single_linear_polymer(const RDKit::ROMol& monomer_mol)
...
+        // reset ring info so it can get recomputed by the appropriate layout
+        // method
+        monomer_mol.getRingInfo()->reset();
+        return false;
+    }
+    for (auto monomer : monomer_mol.atoms()) {
+        if (monomer->getProp<bool>(BRANCH_MONOMER)) {
+            return false;
+        }
+    }
```
rachelnwalker, line 719: I'm a little confused on this because I think double stranded NAs will definitely have branch monomers, I would expect this to always return false what am I missing? The base is always labeled as a branch monomer

rachelnwalker (commented):
Images you included look great! Could you add some double stranded NAs where the 2nd strand needs to be inverted?

src/schrodinger/rdkit_extensions/helm/cg_coordgen.cpp
```diff
@@ -101,6 +101,13 @@ lay_out_chain(RDKit::ROMol& polymer, const RDKit::Atom* start_monomer,
     }
 }
 
+static bool is_nucleic_acid(const RDKit::ROMol& polymer)
+{
+    auto polymerId = polymer.getProp<std::string>(POLYMER_ID);
```
KevKeating, line 106: Nit: should this be `polymer_id` instead of `polymerId`?

src/schrodinger/rdkit_extensions/helm/cg_coordgen.cpp
```diff
@@ -679,11 +686,38 @@ static bool is_single_linear_polymer(const RDKit::ROMol& monomer_mol)
...
+        return false;
+    }
+    for (auto polymer : polymers) {
+        if (!is_nucleic_acid(*polymer)) {
+            return false;
+        }
+    }
+    const auto polymer_bonds =
+        get_bonds_between_polymers(*polymers[0], *polymers[1]);
+    if (polymer_bonds.size() < 1) {
```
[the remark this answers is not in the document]
ZontaNicola: no reason, I don't have a preference so if anybody wants this changed I'm happy to

src/schrodinger/rdkit_extensions/helm/cg_coordgen.cpp
```diff
@@ -706,8 +747,7 @@ unsigned int compute_monomer_mol_coords(RDKit::ROMol& monomer_mol)
             const auto polymer_bonds =
                 get_bonds_between_polymers(*polymer, *last_polymer);
             auto rotate_polymer =
-                !polymer_bonds.empty() &&
-                (polymer_bonds.size() > 1 || is_nucleic_acid(*polymer));
+                !polymer_bonds.empty() && (polymer_bonds.size() > 1);
```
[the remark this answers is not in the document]
ZontaNicola: I think this was an attempt to draw ds nucleic acids, which we now special case

PR #98, opened 2025-10-07
Local Windows builds now succeed

KevKeating:
* Linked Case: BLDMGR-10073
* Branch: main
 
### Description
With this change plus https://github.com/schrodinger/sketcher/pull/97, my local Windows builds now succeed again.  This incorporates some of the changes that @saptarshi-sdgr figured out for https://github.com/schrodinger/sketcher/pull/73 and https://github.com/schrodinger/package-factory_recipes/pull/269, as well as two additional minor changes:

- `-wd4834` (allow `[[nodiscard]]` values to be discarded) is now only applied to the unit tests instead of all files
- the `Q_INIT_RESOURCE` call is only used in static builds, which should avoid the need for explicitly linking `sketcher_autogen/PMZGBCCIHK/qrc_sketcher.cpp.obj` in shared builds

### Testing Done
Local Windows builds succeed

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -76,7 +76,10 @@ function(setup_target TARGET)
   endif()
   # Compilation optimization, warnings, and error handling
   if(MSVC)
-    target_compile_options(${TARGET} PRIVATE -WX)
+    target_compile_options(${TARGET} PRIVATE -WX -EHsc)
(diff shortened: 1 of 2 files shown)


cdvonbargen (approved):
Yes!! This is what I was looking for, thank you!

PR #107, opened 2025-10-11
SKETCH-2577: Fix wiggly bond preservation

cdvonbargen:
* Linked Case: SKETCH-2577
* Branch: 2026-1
 
### Description
Wiggly bonds were being lost during normal molecule operations for several reasons. The wedgeMolBonds() function calls ClearSingleBondDirFlags() which clears the BondDir::UNKNOWN flag that indicates wiggly bonds. Additionally, RDKit's insertMol() doesn't preserve the _MolFileBondCfg property that the sketcher's prepare_mol function checks to detect wiggly bonds. Finally, CXSMILES input sets BondDir::UNKNOWN but doesn't set the _MolFileBondCfg property, creating an inconsistency with MDL format handling.

The fix involves capturing wiggly bond state before it gets cleared and restoring it afterward. In stereochemistry.cpp, the code now saves which bonds are wiggly by checking _MolFileBondStereo (MDL V2000), _MolFileBondCfg (MDL V3000), and BondDir::UNKNOWN (CXSMILES) before calling ClearSingleBondDirFlags(). After the wedging calculation completes, it restores the wiggly bond directions.

In mol_model.cpp, when adding molecules, the code now explicitly copies the BondDir from the source molecule and sets _MolFileBondCfg=2 for CXSMILES-sourced wiggly bonds so that prepare_mol can properly detect them.

### Testing Done
Added test to demonstrate what currently works, as well as what is fixed.
Also added a test to demonstrate the issue with insertMol

### DIFF
--- a/src/schrodinger/rdkit_extensions/stereochemistry.cpp
+++ b/src/schrodinger/rdkit_extensions/stereochemistry.cpp
@@ -75,6 +75,30 @@ void wedgeMolBonds(RDKit::ROMol& mol, const RDKit::Conformer* conf)
     RDKit::ClearSingleBondDirFlags(mol);
     RDKit::Chirality::clearMolBlockWedgingInfo(mol);
 
+    // SKETCH-2577 workaround: hide wiggly bonds from WedgeMolBonds by
+    // temporarily changing them to Bond::OTHER. WedgeMolBonds only considers
+    // Bond::SINGLE bonds, so this prevents pickBondToWedge from stealing a
+    // wiggly bond to express a neighboring chiral atom's stereo — which the
+    // restore step below would then overwrite with UNKNOWN, losing that
+    // atom's wedge representation entirely.
+    //
+    // The upstream fix lives in RDKit (pickBondToWedge skips bonds with
+    // BondDir::UNKNOWN or _UnknownStereo=1). Once that lands in the RDKit
+    // version sketcher uses, this OTHER-type swap can go away — only the
+    // _UnknownStereo → BondDir::UNKNOWN restore below is still needed
+    // (clearSingleBondDirFlags clears UNKNOWN to NONE; we put it back so
+    // MolToCXSmiles emits |w:|).
+    std::vector<RDKit::Bond*> wiggly_bonds;
+    for (auto bond : mol.bonds()) {
+        int unknown_stereo{0};
+        if (bond->getPropIfPresent(RDKit::common_properties::_UnknownStereo,
+                                   unknown_stereo) &&
+            unknown_stereo && bond->getBondType() == RDKit::Bond::SINGLE) {
+            wiggly_bonds.push_back(bond);
+            bond->setBondType(RDKit::Bond::OTHER);
+        }
+    }
+
     try {
         // Temporarily silence RDKit's loggers
         RDLog::LogStateSetter silence_rdkit_logging;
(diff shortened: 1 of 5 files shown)


KevKeating (approved):

src/schrodinger/rdkit_extensions/stereochemistry.cpp
```diff
@@ -138,6 +138,24 @@ void wedgeMolBonds(RDKit::ROMol& mol, const RDKit::Conformer* conf)
...
+    std::vector<bool> had_wiggly_bond;
+    had_wiggly_bond.reserve(mol.getNumBonds());
+    for (auto bond : mol.bonds()) {
+        int wiggly_bond_v2000{0};
+        int wiggly_bond_v3000{0};
+        bond->getPropIfPresent(RDKit::common_properties::_MolFileBondStereo,
+                               wiggly_bond_v2000);
+        bond->getPropIfPresent(RDKit::common_properties::_MolFileBondCfg,
+                               wiggly_bond_v3000);
+        bool is_wiggly = (wiggly_bond_v2000 == 4 || wiggly_bond_v3000 == 2 ||
```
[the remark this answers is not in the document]
cdvonbargen: These values are from the MDL molfile specification, not defined as constants in RDKit -- left comments in the code.

ethan-schrodinger (approved):

src/schrodinger/rdkit_extensions/stereochemistry.cpp
```diff
@@ -154,6 +176,15 @@ void wedgeMolBonds(RDKit::ROMol& mol, const RDKit::Conformer* conf)
...
     }
+
+    // Restore wiggly bond directions that were cleared by
+    // ClearSingleBondDirFlags
+    for (size_t bond_idx = 0; bond_idx < mol.getNumBonds(); ++bond_idx) {
+        if (had_wiggly_bond[bond_idx]) {
+            auto bond = mol.getBondWithIdx(bond_idx);
+            bond->setBondDir(RDKit::Bond::BondDir::UNKNOWN);
+        }
+    }
```
rachelnwalker: @ricrogz @cdvonbargen -- does this also address the issue Ricardo described in this case? https://schrodinger.atlassian.net/browse/SKETCH-2584

rachelnwalker (approved):

src/schrodinger/rdkit_extensions/stereochemistry.cpp
```diff
@@ -138,6 +138,28 @@ void wedgeMolBonds(RDKit::ROMol& mol, const RDKit::Conformer* conf)
...
+                               wiggly_bond_v3000);
+        // MDL V2000 uses _MolFileBondStereo=4 for wiggly bonds
+        // MDL V3000 uses _MolFileBondCfg=2 for wiggly bonds
+        // (These values are from the MDL molfile specification, not defined as
+        // constants in RDKit)
+        bool is_wiggly = (wiggly_bond_v2000 == 4 || wiggly_bond_v3000 == 2 ||
+                          bond->getBondDir() == RDKit::Bond::BondDir::UNKNOWN);
+        had_wiggly_bond.push_back(is_wiggly);
+    }
+
```
ricrogz: I would not do it this way: `WedgeMolBonds()` uses a heuristic to decide which bonds to wedge/dash. Even if it disfavors bonds between 2 chiral atoms, there are still chances that  one of the bonds you pick as wiggly ends up being chosen as a wedge/dash bond for a neighboring chiral atom, which would be overwritten with the wiggly bond when you restore them.

What I would do is store the chiral atom referred to by the wiggly bond, then turn the wiggly into, e.g. a wedge bond (a dash one would work too), and then, once  `WedgeMolBonds()` has done its thing, find out which of the bonds around the chiral atom has been wedged/dashed, and then replace that one with a wiggly.

Of course, all that also taking into account the properties.

cdvonbargen: @ricrogz -- I updated this based on your review, ready for another round. 

I also submitted https://github.com/rdkit/rdkit/pull/9267 as an upstream fix

ricrogz (approved):
Not the ideal solution, but it will probably do. This will keep the exact same bonds being wiggly, when it might be better to allow `WedgeMolBonds()` to choose a better bond to be wiggly (same as it does with dash/wedge bonds).

cdvonbargen:
> Not the ideal solution, but it will probably do. This will keep the exact same bonds being wiggly, when it might be better to allow `WedgeMolBonds()` to choose a better bond to be wiggly (same as it does with dash/wedge bonds).

Posted https://github.com/rdkit/rdkit/pull/9267

PR #130, opened 2025-10-27
SKETCH-2593: fix export macro clash

ricrogz:
Part of SKETCH-2593. When the rdkit_extensions or sketcher headers are included in other Core Suite files, the export macros clash with the ones already defined in the suite. This checks for the existence of the macros before attempting to define them. It also allows overriding them, in case anyone needs to do thath.

### DIFF
--- a/include/schrodinger/rdkit_extensions/definitions.h
+++ b/include/schrodinger/rdkit_extensions/definitions.h
@@ -2,12 +2,20 @@
 
 #pragma once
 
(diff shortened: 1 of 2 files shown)


ethan-schrodinger (approved):

rachelnwalker (approved):
@ricrogz @cdvonbargen I realize this slept through the cracks a while ago - do we still need this as apart of SKETCH-2593?

PR #301, opened 2026-04-07
SKETCH-2483: Implementing monomer click-and-drag

KevKeating:
* Linked Case: SKETCH-2483

### Description
This implements click-and-drag for the monomeric scene tools.  Here are some screen shots:  (Chris, I mostly added you in case you wanted to look at the pretty screen shots.  I don't think there's anything here that critically needs your input, so feel free to ignore everything else.)

A drag started from an existing monomer
image

A drag from empty space to empty space adds two new monomers, which is the same as in atomistic mode
image

A drag started from the monomer uses the default attachment point (if available), which is the C terminus for peptides.  Dragging to another existing monomer defaults to using the N terminus to create a standard backbone bone.
image

This drag started from the X attachment points, and the scene tool is smart enough to default to using the X attachment point at the other end if the user hovers over the monomer.
image

The drags can start and end on either empty space, an existing monomer, or an unbound attachment point of an existing monomer (which appear when the cursor is moved over an existing monomer).  As with the atomistic tools, clicking on empty space and dragging to an empty space will add two atoms/monomers.  Because of this, the new logic ignores click-and-drags from empty spaces when a nucleic acid sugar or phosphate tool is selected, since it doesn't make much biological sense to create a dimer of either of those.

I had Claude write a bunch of integration-level tests that would trigger both clicks and clicks-and-drags by sending events to the Scene.  These would've caught a number of crash bugs that I ran into (and fixed) during development of this PR.  These tests found at least two problems:

- the Scene wasn't updating properly in the test, probably because it hadn't been shown, which was then leading to intermittent crashes.  This wasn't a "real" problem - it was entirely due to the tests themselves.  I have a few ideas for how to potentially fix this.
- the HELM strings that get generated aren't quite right.  I thought that rdkit_extensions would do more sanitization of the molecules than it actually does, so we wind up with weird chain names (e.g. PEPTIDE99999) and disconnected monomers placed into existing chains, which leads to the HELM string showing them as connected.  I'll work on fixing this in a separate PR.  I hadn't previously noticed this issue since the Sketcher GUI doesn't currently include HELM as an available format (which should get fixed as part of SKETCH-2717).

Claude "helpfully" worked around problem two by creating a special function to do HELM string comparisons that would trim down the strings until they matched.  It then "solved" problem one by erasing the body of all of the new tests and replacing them with TODO comments explaining what the test was supposed to do.  It then declared the task done since the test file now passed.  Fortunately, I'd committed the tests to my local git branch before Claude started "fixing" them.  I'll revive the tests and figure out the Scene updating issue once I get the HELM strings fixed.

There are a few additional minor things that need fixing unrelated to what Claude found:
- The cursor hint should disappear when dragging from empty space.  It already disappears at other times when a hint structure is shown (which matches the atomistic ring tool behavior).  I just noticed this issue when making the screen shots above.
- I need to add some additional sanity checking to make sure that the user can't add multiple non-standard connections between two monomers, since our monomer_mol format doesn't allow that.  (Bonds can represent two different monomeric connections, but one of them must be a standard backbone connection.)

This PR contains a fair bit of refactoring so that the click and click-and-drag logic can share as much code as possible.  It took a while to get right, but I actually like the new organization: it very explicitly separates the tasks of figuring out what to draw for the hint, drawing the hint, and committing to MolModel.

### Testing Done

Tested via desktop app on Windows.  The tests that Claude wrote (and then deleted) are definitely helpful, so I'll add those in once the two bugs above are fixed.

### DIFF
--- a/src/schrodinger/sketcher/tool/draw_monomer_scene_tool.h
+++ b/src/schrodinger/sketcher/tool/draw_monomer_scene_tool.h
@@ -38,7 +38,7 @@ namespace sketcher
 
 class MonomerHintFragmentItem;
 class UnboundMonomericAttachmentPointItem;
-class HintFragmentMonomerInfo;
+struct HintFragmentMonomerInfo;
 
 using MonomerAndAPItems =
     std::pair<AbstractMonomerItem*, UnboundMonomericAttachmentPointItem*>;
--- a/src/schrodinger/sketcher/tool/draw_monomer_scene_tool.h
+++ b/src/schrodinger/sketcher/tool/draw_monomer_scene_tool.h
@@ -358,7 +358,7 @@ class SKETCHER_API DrawMonomerSceneTool : public StandardSceneToolBase
 
     /**
      * Determine whether the end of current click-and-drag operation is over an
-     * existing monomer or not.  If not, return the direction of the drag.
+     * existing monomer or not. If not, return the direction of the drag.
      * @param scene_pos The scene coordinates representing the end of the
      * click-and-drag.
      * @return If scene_pos is over an existing monomer, returns a pair of
@@ -380,10 +380,11 @@ class SKETCHER_API DrawMonomerSceneTool : public StandardSceneToolBase
         const HintFragmentMonomerInfo& hint_end_monomer_info);
 
     /**
-     * If the user starting a "valid" click-and-drag, create and return the
-     * HintFragmentMonomerInfo object describing the start of the drag.
-     * Otherwise, return std::nullopt. See the createDragHintIfDragStartValid
-     * docstring for an explanation of valid versus invalid click-and-drags.
+     * If the user has began a "valid" click-and-drag, create and return the
+     * HintFragmentMonomerInfo object describing the starting location/monomer
+     * of the drag. Otherwise, return std::nullopt. See the
+     * createDragHintIfDragStartValid docstring for an explanation of valid
+     * versus invalid click-and-drags.
      */
     std::optional<HintFragmentMonomerInfo>
     getHintFragmentMonomerInfoForDragStart();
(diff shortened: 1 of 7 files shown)


src/schrodinger/sketcher/tool/draw_monomer_scene_tool.h
```diff
@@ -204,6 +289,119 @@ class SKETCHER_API DrawMonomerSceneTool : public StandardSceneToolBase
...
+    std::pair<DragEndInfo, AbstractMonomerItem*>
+    getDragEndInfo(const QPointF& scene_pos);
+
+    void addDragStructureToMolModel(
+        const HintFragmentMonomerInfo& hint_start_monomer_info,
+        const HintFragmentMonomerInfo& hint_end_monomer_info);
+
+    /**
+     * If the user starting a "valid" click-and-drag, create and return the
+     * HintFragmentMonomerInfo object describing the start of the drag.
```
ethan-schrodinger: Am I crazy, or is it a little confusing that we can have "start" or "end" refer both to the time at which the drag operation starts and ends while also referring to the starting and ending points of the drag operation?

Maybe the meaning is clear enough from context.

ethan-schrodinger (commented):
>Claude "helpfully" worked around problem two by creating a special function to do HELM string comparisons that would trim down the strings until they matched. It then "solved" problem one by erasing the body of all of the new tests and replacing them with TODO comments explaining what the test was supposed to do. It then declared the task done since the test file now passed.

Good stuff. :+1:

ethan-schrodinger (approved):

src/schrodinger/sketcher/tool/draw_monomer_scene_tool.h
KevKeating: Yeah, that is a little confusing.  I rephrased it to "If the user has began a "valid" click-and-drag, create and return the HintFragmentMonomerInfo object describing the starting location/monomer of the drag."

PR #302, opened 2026-04-09
SKETCH-2728: Adding fallback font names to generated SVG files

KevKeating:
* Linked Case: SKETCH-2728, SKETCH-2682, LDID8-1488

### Description

@rachelnwalker finally figured out why the images in Live Design's first column aren't using the correct font: those images are rendered as SVGs, which means that the font will only be correct if the user has Arimo installed as a system-wide font.  If not, the web browser doesn't know enough about Arimo to even know that it's sans serif, so it winds up replacing it with a serif font.  To fix this, I've added a list of fallback fonts, including Arial and Helvetica, to the SVG file.  We're only including the font name in the SVG file, not the font itself, so we don't need to worry about font licensing.  With this change, the generated SVGs look correct if I open them up in my web browser, and @juzerzarif confirmed that he sees the same thing.

My first attempt at this was to specify the alternate fonts as QFont families, but QSvgGenerator ignores these.  (I checked in the Qt source code and confirmed that's the expected, but unfortunate, behavior.)  As a result, I added some logic to our SVG post processing to manually add in the additional font names.

I'm assuming that we'll want to add this commit to the release branch in both `sketcher` and `mmshare`.  I don't think I have permissions to merge a PR directly into the release branch on `sketcher`, but I can cherry-pick and manually push after this gets merged into `main`.  When I put up an `mmshare` PR, I'll do that against the release branch so it can go through engineering approval.

### Testing Done
I saved an SVG from Sketcher and confirmed that it looks correct.  I also updated the SVG unit test to include a check for the font list.

### DIFF
--- a/src/schrodinger/sketcher/image_generation.cpp
+++ b/src/schrodinger/sketcher/image_generation.cpp
@@ -383,6 +383,43 @@ QImage get_qimage(const T& input, const RenderOptions& opts)
     return *dynamic_cast<QImage*>(paint_device.get());
 }
 
+/**
+ * Post-process the contents of the SVG file after Qt has generated it
+ * @param svg_data The contents of the SVG file
+ * @param opts The options used to render the SVG
+ */
+void post_process_svg(QByteArray& svg_data, const RenderOptions& opts)
+{
+    // Qt defines svg size in mm, but our code needs it specified in pixels.
+    // This is a hack to make sure we get the right size definition.
+    auto start_of_size = svg_data.indexOf("svg width");
+    auto end_of_size = svg_data.indexOf(" viewBox", start_of_size);
+    if (start_of_size == -1 || end_of_size == -1) {
+        throw std::runtime_error("Failed to find svg size definition");
+    }
+    svg_data.remove(start_of_size, end_of_size - start_of_size);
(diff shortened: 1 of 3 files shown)


juzerzarif (approved):

rachelnwalker (approved):

src/schrodinger/sketcher/image_generation.cpp
```diff
@@ -383,6 +383,43 @@ QImage get_qimage(const T& input, const RenderOptions& opts)
     return *dynamic_cast<QImage*>(paint_device.get());
 }
 
+/**
+ * Post-process the contents of the SVG file after Qt has generated it
+ * @param svg_data The contents of the SVG file
+ * @param opts The options used to render the SVG
```
ethan-schrodinger, line 389: Would it make sense to include here the goals of the post-processing?

ethan-schrodinger (approved):

src/schrodinger/sketcher/image_generation.cpp
KevKeating: I thought about doing that, but it seemed like any description here would become stale far more quickly than the comments that are above each block of code in the body of the function.

PR #303, opened 2026-04-13
SKETCH-2483: Fixing HELM string output for drawn structures and adding integration tests

KevKeating:
* Linked Case: SKETCH-2483

There are two main changes here:

- I've added a bunch of integration tests for the monomer scene tool.  These tests mimic clicks and click-and-drags and then confirm that the generated HELM string matches the expected value. I had Claude write the initial version of these tests but almost all of them crashed, which Claude "helpfully" fixed by replacing all of the crashing tests with TODOs. I pulled the deleted test bodies out of my git history and wound up rewriting almost the entirety of the file.  The tests now all pass, and the test fixture is rich enough that it's fairly trivial to add new scenarios.
- I fixed some of the real bugs that Claude managed to uncover before it replaced all of the tests with TODOs These bugs related to HELM string generation: Monomers are now properly assigned to chains if they share a standard backbone bond, and chain number now starts at 1 instead of consistently using 99999.  I also added some MolModel unit tests that reproduced these failures, since the issues were in the data that MolModel was passing to monomer_mol.

### Testing Done
Added new unit tests and confirmed that they passed.

### DIFF
--- a/src/schrodinger/sketcher/rdkit/monomeric.cpp
+++ b/src/schrodinger/sketcher/rdkit/monomeric.cpp
@@ -4,6 +4,7 @@
 #include <functional>
 #include <numeric>
 #include <optional>
+#include <string>
 
 #include <boost/range/join.hpp>
 
@@ -783,6 +784,44 @@ void merge_chains(RDKit::ROMol& mol, const std::string_view merge_from,
         res_info->setChainId(merge_to);
         res_info->setResidueNumber(++new_res_num);
     }
+}
+
+int get_chain_num(const std::string_view chain_name,
+                  const rdkit_extensions::ChainType chain_type)
+{
+    auto prefix = rdkit_extensions::toString(chain_type);
+    auto first_num_char = prefix.size();
+    if (chain_name.substr(0, first_num_char) != prefix) {
+        return -1;
+    }
+    auto chain_num_text = chain_name.substr(first_num_char);
+    try {
+        return std::stoi(std::string(chain_num_text));
+    } catch (const std::invalid_argument&) {
+        return -1;
+    } catch (const std::out_of_range&) {
+        return -1;
+    }
+}
+
+std::string
+get_first_available_chain_name(const RDKit::ROMol& mol,
+                               const rdkit_extensions::ChainType chain_type)
+{
+    auto prefix = rdkit_extensions::toString(chain_type);
+    std::unordered_set<int> chain_nums;
+    for (auto chain_name : rdkit_extensions::get_polymer_ids(mol)) {
+        if (!chain_name.starts_with(prefix)) {
+            continue;
+        }
+        int cur_chain_num = get_chain_num(chain_name, chain_type);
+        chain_nums.insert(cur_chain_num);
+    }
+    int missing_num = 1;
+    while (chain_nums.contains(missing_num)) {
+        ++missing_num;
+    }
+    return fmt::format("{}{}", prefix, missing_num);
 }
 
 } // namespace sketcher
--- a/src/schrodinger/rdkit_extensions/helm/monomer_coordgen.cpp
+++ b/src/schrodinger/rdkit_extensions/helm/monomer_coordgen.cpp
@@ -2719,10 +2719,14 @@ inline RDGeom::Point3D get_monomer_size(const RDKit::ROMol& mol,
                                         unsigned int index)
 {
     RDGeom::Point3D size(MONOMER_MINIMUM_SIZE, MONOMER_MINIMUM_SIZE, 0);
-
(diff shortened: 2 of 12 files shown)


src/schrodinger/sketcher/rdkit/monomeric.cpp
```diff
@@ -785,5 +786,43 @@ void merge_chains(RDKit::ROMol& mol, const std::string_view merge_from,
...
+    if (chain_name.substr(0, first_num_char) != prefix) {
+        return -1;
+    }
+    auto chain_num_text = chain_name.substr(first_num_char);
+    try {
+        return std::stoi(std::string(chain_num_text));
+    } catch (const std::invalid_argument&) {
+        return -1;
+    } catch (const std::out_of_range&) {
+        return -1;
```
ethan-schrodinger, line 803: Is there any place where the calling code processes a `-1` return value? In other words, are we doing ourselves any favors by prevent an exception here?

ethan-schrodinger (approved):

src/schrodinger/sketcher/rdkit/monomeric.cpp
KevKeating: There's nothing that explicitly looks for a `-1`, but `get_first_available_chain` would be able to handle a `-1`.  It would effectively cause `get_first_available_chain` to ignore that chain since the `-1` would get inserted into the `chain_nums` set but never checked.  Ignoring the chain seems like the right behavior if the chain name doesn't follow the expected format.  In theory, someone could call `addRDKitMolecule` on an embedded `SketcherWidget` and pass in a molecule with malformed chain names.  I'm not sure we need to go out of our way to support that behavior, but there's no reason to go out of our way to crash if that happens.

KevKeating (commented):
I've got a fix for the memtest failure that I'll post soon.  It's a real memory leak that's coming from `getHintFragmentMonomerInfoForDragStart` and `getHintFragmentMonomerInfoForDragEnd`.  When those methods are called from `onLeftButtonDragMove`, the new `RDKit::Atom*`s that they create are added to a `MonomerHintFragmentItem` and destroyed when that graphics item is destroyed.  However, those methods are also called from `onLeftButtonDragRelease` so the atoms can be added to MolModel, but MolModel makes a copy of the Atoms instead of taking ownership, meaning that the Atoms leak.

The Mac test failures appear to be a real crash, possibly coming from monomer_coordgen?  Maybe?  Tracking it down has been a pain since I don't have a Mac.  If Claude can't make progress on it soon, I may skip the tests on Mac and file a spin off case.  Assuming it's a real crash that also happens in the GUI, someone with a Mac will hopefully have a much easier time with it.

KevKeating:
The latest commit fixes both the memtest failures and the Mac failures:

My previous comment explained the cause of the memtest failures
>  It's a real memory leak that's coming from `getHintFragmentMonomerInfoForDragStart` and `getHintFragmentMonomerInfoForDragEnd`.  When those methods are called from `onLeftButtonDragMove`, the new `RDKit::Atom*`s that they create are added to a `MonomerHintFragmentItem` and destroyed when that graphics item is destroyed.  However, those methods are also called from `onLeftButtonDragRelease` so the atoms can be added to MolModel, but MolModel makes a copy of the Atoms instead of taking ownership, meaning that the Atoms leak.

To fix it, I changed `HintFragmentMonomerInfo::monomer` from a raw pointer to a `std::unique_ptr`.  As a result, the Atoms will always be destroyed unless something calls `.release()`, which `onLeftButtonDragMove` now does.


The Mac failures turned out to be very strange.  I think it's an RDKit bug?  The `std::bad_any_cast` was coming from [this](https://github.com/schrodinger/sketcher/blob/2693993f9e918e31adaac17b70b1563c34091f58/src/schrodinger/rdkit_extensions/helm/monomer_coordgen.cpp#L2723) getPropIfPresent call in get_monomer_size:
```c++
mol.getAtomWithIdx(index)->getPropIfPresent<RDGeom::Point3D>(
    MONOMER_ITEM_SIZE, size);
```
`std::bad_any_cast` should mean that something is trying to convert a `std::any` value to the wrong type.  (`std::any` is basically an STL version of `QVariant`.)  However, `MONOMER_ITEM_SIZE` is only ever set in that same file, and it's set to be a `RDGeom::Point3D`, so I'm not sure why trying to convert it to the correct type would result in an exception, nor why it would only happen on Mac.  For now, I've wrapped the call with a try...catch that returns the default size value when an exception is caught.  I'm guessing that I should file a SHARED ticket for this?

rachelnwalker (approved):

JarrettSJohnson (approved):
LGTM.

(Told Kevin I'll give the exception a look on my macbook--but I can do that after this is merged)

KevKeating:
I just filed SHARED-12285 for the std::bad_any_cast issue

PR #304, opened 2026-04-13
SKETCH-2727: Fix sketcher freeze when erasing multiple plus signs at once

cdvonbargen:
* Linked Case: SKETCH-2727

### Description
Erasing from `m_pluses` (a `std::vector`) in `removeNonMolecularObjectCommandFunc` invalidates all pointers stored in `m_tag_to_non_molecular_object` for elements at or after the erased position. When `removeCommandFunc` looped over multiple plus tags (e.g. marquee-select-all-then-erase), the second removal dereferenced a dangling pointer, causing heap corruption and a crash.

The fix calls `updateNonMolecularMetadata()` after each erase from `m_pluses` to rebuild the tag→pointer map with valid pointers before the next iteration.

### Testing Done
Added a test that crashes before the change and doesnt after

### DIFF
--- a/src/schrodinger/sketcher/model/mol_model.cpp
+++ b/src/schrodinger/sketcher/model/mol_model.cpp
@@ -2782,6 +2782,9 @@ bool MolModel::removeNonMolecularObjectCommandFunc(
                              return &cur_plus == obj;
                          });
         m_pluses.erase(iter_to_erase);
+        // Rebuild the tag→pointer map: erasing from a vector invalidates
+        // pointers to all subsequent elements.
+        updateNonMolecularMetadata();
     }
     return selected;
 }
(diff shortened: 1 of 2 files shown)


KevKeating (approved):

src/schrodinger/sketcher/model/mol_model.cpp
```diff
@@ -2782,6 +2782,9 @@ bool MolModel::removeNonMolecularObjectCommandFunc(
                              return &cur_plus == obj;
                          });
         m_pluses.erase(iter_to_erase);
+        // Rebuild the tag→pointer map: erasing from a vector invalidates
+        // pointers to all subsequent elements.
+        updateNonMolecularMetadata();
```
ethan-schrodinger, line 2787: Hmm. Should we consider using a `std::list` rather than a `std::vector`? I don't think lists have this "gotcha".

ethan-schrodinger (approved):

src/schrodinger/sketcher/model/mol_model.cpp
cdvonbargen: That's right and would work, but I let claude churn and it pushed back:
```
 Changing MolModel::m_pluses to std::list would break this assignment and require converting between types in takeSnapshot() and restoreSnapshot(). That's extra churn for questionable benefit.
```
So I'd have to make deeper changes. I think this small fix is likely the best way forward here.

PR #305, opened 2026-04-16
SKETCH-2483: Fixing minor issues with the monomer scene tool

KevKeating:
* Linked Case: SKETCH-2483

### Description
This fixes a handful of minor issues with the monomeric scene tool:
- Previously, hovering over a connector would label the attachment points at both ends of the connector, but I broke that at some point.  This branch fixes that so it works again.
- The cursor hint is now hidden when doing a click-and-drag from empty space.  The cursor hint had already been hidden in other scenarios where we draw a hint structure, such as click-and-drag from an existing monomer and hovering over a monomer.
- Click-and-dragging between two existing monomers will no longer attempt to add multiple standard or multiple custom connections between a pair of monomers.  The monomer mol format only allows one connection of each type between any given monomer pair, so attempting to add an additional connection would have failed.
- MolModel no longer assigns negative residue numbers to new residues.  RDKit stores the residue numbers as signed ints, but `rdkit_extensions::get_residue_number` returns them as unsigned ints, so negative numbers seem like a bad idea.
- MolModel now ensures that residue numbers for new residues are unique.  I'm actually not entirely sure whether this is necessary or not.  I was running into some issues with cleanup that I thought were caused by duplicate residue numbers, but this change didn't fix them, so my new guess is that the issues are somehow related to SKETCH-2716 (i.e. clean up on monomer mols currently seems to have problems in general).  Regardless, there's not a ton of additional logic required for this, and I could envision duplicate residue numbers causing some sort of issue in edge cases, so it seems worth leaving in for now.

I think that this PR finally finishes the monomer scene tool (outside of the crash bug that Jarrett found and has a fix for).  After this, I'll probably start working on the scene tool for adding a whole nucleotide at once.

### Testing Done
Tested the behavior on the Windows desktop app.  I also added some new integration tests to make sure that click-and-drag won't try to add too many connections to an existing Bond.

### DIFF
--- a/src/schrodinger/sketcher/model/mol_model.cpp
+++ b/src/schrodinger/sketcher/model/mol_model.cpp
@@ -683,8 +683,8 @@ void MolModel::addMonomer(const std::string_view res_name,
 }
 
 static MonomerType
-monomer_type_from_chain_type(const rdkit_extensions::ChainType chain_type,
-                             const std::string_view res_name)
+get_monomer_type_from_chain_type(const rdkit_extensions::ChainType chain_type,
+                                 const std::string_view res_name)
 {
     using rdkit_extensions::ChainType;
     switch (chain_type) {
@@ -745,7 +745,7 @@ get_residue_number_for_new_monomer(const std::string_view res_name,
 {
     using rdkit_extensions::ChainType;
 
-    auto monomer_type = monomer_type_from_chain_type(chain_type, res_name);
+    auto monomer_type = get_monomer_type_from_chain_type(chain_type, res_name);
     auto existing_res_nums = get_all_residue_numbers_of_monomer_type_in_polymer(
         monomer_type, bound_to_monomer);
 
--- a/src/schrodinger/sketcher/model/mol_model.cpp
+++ b/src/schrodinger/sketcher/model/mol_model.cpp
@@ -682,18 +682,60 @@ void MolModel::addMonomer(const std::string_view res_name,
     doCommandUsingSnapshots(cmd_func, "Add monomer", WhatChanged::MOLECULE);
 }
 
+static MonomerType
+monomer_type_from_chain_type(const rdkit_extensions::ChainType chain_type,
+                             const std::string_view res_name)
+{
+    using rdkit_extensions::ChainType;
+    switch (chain_type) {
+        case (ChainType::PEPTIDE):
+            return MonomerType::PEPTIDE;
+        case (ChainType::CHEM):
+            return MonomerType::CHEM;
+        default:
+            return get_na_monomer_type_from_res_name(res_name);
+    }
+}
+
+/**
+ * @return a set of all residue numbers for monomers of the specified type that
+ * are in the same polymer as the given atom.
+ */
+static std::unordered_set<int>
+get_all_residue_numbers_of_monomer_type_in_polymer(
+    const MonomerType monomer_type, const RDKit::Atom* const atom_in_polymer)
+{
+    // we make a copy of the molecule so we can flag it as monomeric
+    auto mol = atom_in_polymer->getOwningMol();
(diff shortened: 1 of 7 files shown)


src/schrodinger/sketcher/model/mol_model.cpp
```diff
@@ -726,84 +772,13 @@ get_residue_number_for_new_monomer(const std::string_view res_name,
...
-    auto ap_num_one = ap_name_to_num(ap_name_one);
-    auto ap_num_two = ap_name_to_num(ap_name_two);
-    bool flip = ap_num_one < ap_num_two;
-    std::string_view start_ap_name = ap_name_one;
-    std::string_view end_ap_name = ap_name_two;
-    if (flip) {
-        std::swap(start_ap_name, end_ap_name);
-    }
-    std::string linkage = fmt::format("{}-{}", start_ap_name, end_ap_name);
-    return {linkage, flip};
```
KevKeating, line 806: Moved to monomeric.cpp/.h

src/schrodinger/sketcher/rdkit/monomeric.cpp
```diff
@@ -824,5 +824,80 @@ get_first_available_chain_name(const RDKit::ROMol& mol,
...
+                        const rdkit_extensions::ChainType chain_type_one,
+                        const RDKit::Atom* const monomer_two,
+                        const std::string_view linkage)
+{
+    auto monomer_two_res_name = get_monomer_res_name(monomer_two);
+    auto monomer_two_chain_type = rdkit_extensions::getChainType(*monomer_two);
+    return get_is_custom_bond(res_name_one, chain_type_one,
+                              monomer_two_res_name, monomer_two_chain_type,
+                              linkage);
+}
```
KevKeating, line 900: Moved from mol_model.cpp with very minor changes (get_is_custom_bond now has three overloads instead of two)

src/schrodinger/sketcher/rdkit/monomeric.h
```diff
@@ -263,5 +263,49 @@ SKETCHER_API std::string
...
+SKETCHER_API bool get_is_custom_bond(const RDKit::Atom* const monomer_one,
+                                     const RDKit::Atom* const monomer_two,
+                                     const std::string_view linkage);
+
+/// @overload
+SKETCHER_API bool
+get_is_custom_bond(const std::string_view res_name_one,
+                   const rdkit_extensions::ChainType chain_type_one,
+                   const RDKit::Atom* const monomer_two,
+                   const std::string_view linkage);
```
KevKeating, line 308: Moved from mol_model.cpp with very minor changes (get_is_custom_bond now has three overloads instead of two, and build_linkage_string has better documentation for the return values)

src/schrodinger/sketcher/model/mol_model.cpp
KevKeating: This whole big block was moved, not just the last few lines.  I'm not sure why GitHub decided to only show a few of them here.

src/schrodinger/sketcher/model/mol_model.cpp
```diff
@@ -682,18 +682,60 @@ void MolModel::addMonomer(const std::string_view res_name,
     doCommandUsingSnapshots(cmd_func, "Add monomer", WhatChanged::MOLECULE);
 }
 
+static MonomerType
+monomer_type_from_chain_type(const rdkit_extensions::ChainType chain_type,
```
ethan-schrodinger: Nit: `get_monomer_type_from_chain_type()`?

ethan-schrodinger (approved):
LGTM

src/schrodinger/sketcher/model/mol_model.cpp
KevKeating: Changed.

PR #306, opened 2026-04-16
SKETCH-2759: Preserve crossed double-bond marker from SDF/molblock paste

cdvonbargen:
* Linked Case: SKETCH-2759

### Description

Pasting an MDL/SDF molblock with a double bond explicitly marked as crossed (V2000 stereo flag 3, V3000 `CFG=2`) silently resolved to E or Z from the 2D layout, dropping the user's "unspecified" intent.

The clobber happens in `assign_stereochemistry_with_bond_directions_and_coordinates`: it clears `Bond::Stereo` unconditionally and only restores `BondDir` after `cleanIt=true` perception. By the time we get here, `to_rdkit` has already cleared `BondDir` to `NONE` while preserving `Stereo=STEREOANY`, so the snapshot misses the crossed marker entirely. Fix: also snapshot `Stereo==STEREOANY || BondDir==EITHERDOUBLE` on entry, and restore the canonical pair after perception.

Rescoped from SKETCH-2722 (bare `=` in SMILES), which is being closed wont-fix — ketcher and marvin resolve unspecified SMILES double bonds the same way the sketcher currently does, so SDF input is the only case with unambiguous user intent.

### Testing Done

Added two tests in `test_stereochemistry.cpp` exercising the full `to_rdkit` → `update_molecule_on_change` path for both V3000 and V2000 crossed-bond inputs.

Co-Authored-By: Claude Opus 4.7 <noreply@anthropic.com>


ethan-schrodinger (approved):

src/schrodinger/rdkit_extensions/stereochemistry.cpp
```diff
@@ -56,6 +56,22 @@ void assign_stereochemistry(RDKit::ROMol& mol)
...
+    // inputs whose coords resolve the stereo (MDL, 2D molblocks),
+    // findPotentialStereo would otherwise return Unspecified and we'd
+    // incorrectly tag a bond that the layout already disambiguates.
+    for (const auto& si :
+         RDKit::Chirality::findPotentialStereo(mol, /*cleanIt=*/false)) {
+        if (si.type == RDKit::Chirality::StereoType::Bond_Double &&
+            si.specified == RDKit::Chirality::StereoSpecified::Unspecified) {
+            mol.getBondWithIdx(si.centeredOn)
+                ->setBondDir(RDKit::Bond::BondDir::EITHERDOUBLE);
+        }
```
d-b-w: I think this should _also_ `setStereo(STEREOANY)` - they are usually paired.

d-b-w (changes requested):

test/schrodinger/sketcher/rdkit/test_stereochemistry.cpp
```diff
@@ -101,5 +103,76 @@ BOOST_AUTO_TEST_CASE(test_wedgeMolBonds)
...
+        V2000_CROSSED, rdkit_extensions::Format::MDL_MOLV2000);
+    update_molecule_on_change(*mol);
+
+    auto* bond = mol->getBondBetweenAtoms(1, 2);
+    BOOST_REQUIRE(bond != nullptr);
+    BOOST_TEST(bond->getStereo() == RDKit::Bond::BondStereo::STEREOANY);
+    BOOST_TEST(bond->getBondDir() == RDKit::Bond::BondDir::EITHERDOUBLE);
+}
+
 } // namespace sketcher
```
d-b-w, line 203: Is there already a test that ensures that this doesn't break SMILES?

d-b-w (changes requested):
I approve _if_ there is already a test that this doesn't break SMILES.

test/schrodinger/sketcher/rdkit/test_stereochemistry.cpp
cdvonbargen: added one

test/schrodinger/sketcher/rdkit/test_stereochemistry.cpp
```diff
@@ -101,5 +103,89 @@ BOOST_AUTO_TEST_CASE(test_wedgeMolBonds)
...
+        V2000_CROSSED, rdkit_extensions::Format::MDL_MOLV2000);
+    update_molecule_on_change(*mol);
+
+    auto* bond = mol->getBondBetweenAtoms(1, 2);
+    BOOST_REQUIRE(bond != nullptr);
+    BOOST_TEST(bond->getStereo() == RDKit::Bond::BondStereo::STEREOANY);
+    BOOST_TEST(bond->getBondDir() == RDKit::Bond::BondDir::EITHERDOUBLE);
+}
+
+BOOST_AUTO_TEST_CASE(test_smiles_specified_stereo_not_clobbered)
```
d-b-w, line 177: We should also have a test of the _opposite_ of SKETCH-2722 -

paste in `CC=CC` should resolve.

d-b-w (changes requested):

rachelnwalker (approved):

d-b-w (approved):
thanks!

PR #307, opened 2026-04-17
SKETCH-2483: Fixing crash when creating non-base-pair connections with the pair attachment point

KevKeating:
* Linked Case: SKETCH-2483

### Description
I ran into a crash with click-and-drag when I was starting to work on the full nucleotide tools, so I figured I'd put up the fix.  The underlying problem here is that the `addConnection` function in `monomer_mol` assumes that the "pair" attachment point on nucleic acid bases is only ever connected to another "pair" attachment point.  That's a reasonable assumption for molecules created from standard-ish HELM strings, but the new click-and-drag functionality makes it possible to connect this attachment point to pretty much anything, particularly mid-drag.  This was leading to a crash when `addConnection` used a `linkage.front() != 'p'` check to mean "this `linkage` string must not involve any pair attachment points" and causing a downstream unhandled exception when trying to parse attachment point numbers from strings along the lines of "R3-pair".  This diff modifies that check to just call `getAttchpts`, which parses attachment point numbers, and interpret any exceptions it throws as "this linkage string involves attachment points that don't follow the R<#> naming convention, so we don't need to worry about assigning directionality to the bond".

As part of this PR, I also modified `DrawMonomerSceneTool::createHintFragmentItem` to correctly specify `is_custom_bond` when calling `addConnection`.  I'm not sure if failing to specify this actually had any noticeable effects in this particular scenario (since it's only a hint structure that will always involve exactly two monomers and a single bond), but there's no reason not to fix this.

### Testing Done
Tested via the Windows desktop app.  Also added two new unit/integration tests involving the "pair" attachment point.


src/schrodinger/rdkit_extensions/monomer_mol.cpp
```diff
@@ -384,20 +384,23 @@ addConnection(RDKit::RWMol& monomer_mol, size_t monomer1, size_t monomer2,
...
-            }
+        unsigned int begin_attchpt = 0;
+        unsigned int end_attchpt = 0;
+        try {
+            std::tie(begin_attchpt, end_attchpt) = getAttchpts(linkage);
+        } catch (...) {
+            // if we can't parse the linkage (e.g. "pair-pair"), assume it's
+            // non-directional
+        }
+        bool set_directional_bond = begin_attchpt != end_attchpt;
```
ethan-schrodinger: Nit: do we need a variable for this? It looks like it's only used in one place now.

ethan-schrodinger (approved):

src/schrodinger/rdkit_extensions/monomer_mol.cpp
rachelnwalker: yeah, I think line 406 could just be the `else`

rachelnwalker (commented):
I'm not convinced `R3-pair` should ever be a valid connection - I'd think `pair-pair` only for hbonds?

KevKeating:
> I'm not convinced `R3-pair` should ever be a valid connection - I'd think `pair-pair` only for hbonds?

How would we indicate a connection between a nucleic acid base and a peptide side chain?

rachelnwalker:
> > I'm not convinced `R3-pair` should ever be a valid connection - I'd think `pair-pair` only for hbonds?
> 
> How would we indicate a connection between a nucleic acid base and a peptide side chain?

Pinging @emilyrrhodes on this: my non-scientist perception says we shouldn't allow this, but curious what Emily thinks

JarrettSJohnson:
> I'm not convinced `R3-pair` should ever be a valid connection - I'd think `pair-pair` only for hbonds?

Would sketcher ever be needed to sketch out covalent peptide-nucleic acid crosslinks? Or is that out-of-scope?

KevKeating:
After chatting with Rachel and Emily offline, I've changed how this is handled.  Now, creating a connection that involves at least one "pair" attachment point will create a "pair-pair" linkage, regardless of the identity of the other attachment point.  Emily pointed out that the "pair" attachment point implies hydrogen bonding, while the "R<#>" attachment points imply covalent (or disulfide) bonding, so combining both into a single linkage is nonsensical.

The interface for this probably isn't 100% ideal.  The hint bond will show "pair", but the drag still makes it look like the hovered non-pair attachment point is relevant.  This seems like a minor issue, though, so I think it's fine for now.

Also, I'm not sure if the H-bond versus non-H-bond distinction is something we want to worry about for non-cysteine side chain - side chain interactions.  Presumably those shouldn't be "R3-R3" if they're hydrogen bonds.

> Would sketcher ever be needed to sketch out covalent peptide-nucleic acid crosslinks? Or is that out-of-scope?

It's probably out-of-scope for at least the initial version, but we may want to worry about allowing covalent bonds from bases and hydrogen bonds from side chains at some point, since both of those seem like they could be relevant.

KevKeating:
Also, the latest commit creates some new helper functions: `set_all_atoms_monomeric` and `add_monomer_connection`.  These help make this change a bit more readable, although the main reason I added them is because they'll also be used in SKETCH-2618.

ethan-schrodinger (approved):
Nice.

PR #312, opened 2026-04-23
SKETCH-2735: Make ATOMISTIC_OR_MONOMERIC the default

cdvonbargen:
* Linked Case: SKETCH-2735

### Description

At this point, I think the monomer representation is all there -- Cindy has SOW deliverables for "Q2 monomer sketcher copy/paste to LR", and @rachelnwalker would like both internal and external folks to be able to test monomer coordgen changes more easily.

I've walked this back a bit from the original PR; here's my plan:
- this sets the new default to ATOMISTIC_OR_MONOMERIC and removes the executable/WASM toggle (and per the discussion in standup, *disables* rather than hides the UI elements)
- I'll follow this PR with updating the mmshare sip wrapped sketchers to raise if a user tries to extract a Structure (or mol) when there are monomers present in the sketcher
- Will the evaluate if we still want InterfaceType (and if not, I'll kill it)

As Jarrett pointed out in standup, Im worried about the implications of supporting these 3 modes going forward for the sake of maestro embedded sketchers. Im gonna keep pushing on a single mode, but this seems like a reasonable first step to exploring that.

### Testing Done


include/schrodinger/sketcher/sketcher_widget.h
ethan-schrodinger, line 133: We have multiple modes that are mutually exclusive. Rather than implement independent boolean states for them, wouldn't it be better to have a single variable for tracking state that uses an enum?

src/schrodinger/sketcher/sketcher_widget.cpp
KevKeating, line 1046: You should add a check here for monomer view-only mode, since otherwise it's possible to switch to a monomeric editing tool by pasting a HELM string and then using the keyboard shortcuts.

include/schrodinger/sketcher/sketcher_widget.h
```diff
@@ -71,9 +71,7 @@ class SKETCHER_API SketcherWidget : public QWidget
     Q_OBJECT
 
   public:
-    SketcherWidget(
-        QWidget* parent = nullptr,
-        const InterfaceTypeType interface_type = InterfaceType::ATOMISTIC);
```
KevKeating, line 77: It might still be useful to keep the interface_type setting for embedded Sketchers.  Most panels probably don't want the user to be able to paste a HELM string if the panel is intended for atomistic models.

KevKeating (commented):
It looks like you need to update the `sketcher_allow_monomeric` calls in the Playwright tests.

include/schrodinger/sketcher/sketcher_widget.h
cdvonbargen: `setMonomerViewOnly` is effectively a feature flag, it's just a bridge to turning on monomers on for everyone for everything eventually. It's also not mutually exclusive to the select mode. I'm gonna leave it as a bool for now, but the goal is to kill this off once Kevin & Jarrett are happy with monomer sketcher

include/schrodinger/sketcher/sketcher_widget.h
cdvonbargen: I thought about this, but (a) dont think anyone in Maestro is pasting HELM strings, and (b) hate all the special sketcher modes that have spawned all over maestro. My instinct is that this is acceptable, but we can also ask Sean to poke hard at the various sketchers to make sure the embedded sketchers aren't broken.

Also happy to talk through this with the group during standup

JarrettSJohnson (approved):

ethan-schrodinger (approved):

rachelnwalker (commented):
exciting!

ethan-schrodinger (approved):

.github/workflows/update-wasm-snapshots.yml
```diff
@@ -0,0 +1,67 @@
+name: Update Wasm Snapshots
+
+# Downloads the wasm artifact from a previous "Sketcher Build and Test" run,
+# regenerates the playwright snapshots against it, and uploads the new
```
KevKeating: It took me a few minute to realize that "snapshots" here meant screen shots, not a specific build of Sketcher.  Adding something like "(i.e. screen shots)" after "snapshots" might help?  Maybe also add something like "This action is needed when a commit is expected to change the appearance of the Sketcher." after this first sentence?

KevKeating (approved):

JarrettSJohnson (approved):

PR #313, opened 2026-04-24
SHARED-12308: Add a smiles_to_helm utility

i-tub:
* Linked Case: SHARED-12308

This is pretty rudimentary: reads one SMILES per line from stdin, writes one HELM per line to stdout. Supports optional titles after a space, and specifying a custom monomer database as a command-line argument.

### Testing Done

Added some unit tests that run the program as a subprocess; also played with it a bit from the command line.


test/python/test_smiles_to_helm.py
```diff
@@ -0,0 +1,97 @@
...
+        else:
+            os.environ[DB_ENV_VAR] = old_db
+
+
+def smiles_to_helm(smiles: str, db: str | None = None) -> str:
+    """
+    Helper function to run smiles_to_helm with the given input and optional
+    monomer database file.
+
+    :param smiles: stding to feed into smiles_to_helm
```
i-tub: `stding` was an accidental portmanteau of `stdin` and `string`. 🙄

test/python/test_smiles_to_helm.py
```diff
@@ -0,0 +1,97 @@
...
+    p = subprocess.Popen(
+        cmd,
+        stdin=subprocess.PIPE,
+        stdout=subprocess.PIPE,
+        stderr=subprocess.PIPE,
+        encoding='utf8',
+    )
+    helm_out, err = p.communicate(smiles)
+    if p.returncode != 0:
+        print(os.getcwd())
```
i-tub: Oops, left a debug print.

src/utils/smiles_to_helm.cpp
```diff
@@ -0,0 +1,157 @@
...
+Reads SMILES strings from stdin (one per line, optionally followed by a space
+and a name), converts them to monomeric representation, and outputs HELM
+format to stdout (with optional name).
+
+Options:
+  --db FILE     Load custom monomer database from FILE (.sqlite/.db or .json)
+                WARNING: using a JSON file will store the monomers permanently
+                in the user's database, by default at
+                ~/.schrodinger/custom_monomerlib.db, or at the path defined by
+                the SCHRODINGER_CUSTOM_MONOMER_DB_PATH environment varable.
```
i-tub: I just realized that rather than warning, maybe what we should do here is what the unit tests do, which is to use a temporary database. I think that would best follow the principle of least surprise.

rachelnwalker (approved):
Looks great. Maybe we should add a readme that references steps to build sketcher and the python `smiles_to_helm` function you added?

test/testfiles/X1.json
```diff
@@ -0,0 +1,13 @@
+[
```
cdvonbargen, line 1: can all the testfiles go into test/testfiles/?

CMakeLists.txt
```diff
@@ -403,7 +418,8 @@ install(DIRECTORY ${PROJECT_SOURCE_DIR}/include/schrodinger
 install(FILES ${PROJECT_BINARY_DIR}/include/schrodinger/sketcher/version.h
         DESTINATION ${CMAKE_INSTALL_INCLUDEDIR}/schrodinger/sketcher)
 # Use FILES/PROGRAMS instead of TARGETS to preserve rpaths on Mac
-install(PROGRAMS $<TARGET_FILE:${APP_TARGET}> DESTINATION ${CMAKE_INSTALL_BINDIR})
+install(PROGRAMS $<TARGET_FILE:${APP_TARGET}> $<TARGET_FILE:${SMILES_TO_HELM_TARGET}>
+        DESTINATION ${CMAKE_INSTALL_BINDIR})
```
cdvonbargen: do we want to install this? I think Im just trying to puzzle out/when this goes out to customers

cdvonbargen (approved):

test/testfiles/X1.json
i-tub: Done!

CMakeLists.txt
i-tub: TBH I'm not sure what the implications of installing are, so I'm agnostic. At this point I think all we want is for the binary to be in the build artifacts which are created from the sketcher_app/ directory; installation just copies it to the bin/ directory, right?

PR #314, opened 2026-04-27
SKETCH-2648: Foundation for monomer context menu

JarrettSJohnson:
* Linked Case: SKETCH-2648

### Description
Adds a `MonomerContextMenu` which just initially exposes the `Delete` function. This context menu replaces the prior behavior of showing the atomistic version.

### Testing Done
Tests that `deleteRequested` is emitted.

### DIFF
--- a/src/schrodinger/sketcher/sketcher_widget.cpp
+++ b/src/schrodinger/sketcher/sketcher_widget.cpp
@@ -37,6 +37,7 @@
 #include "schrodinger/sketcher/menu/background_context_menu.h"
 #include "schrodinger/sketcher/menu/bond_context_menu.h"
 #include "schrodinger/sketcher/menu/bracket_subgroup_context_menu.h"
+#include "schrodinger/sketcher/menu/monomer_context_menu.h"
 #include "schrodinger/sketcher/menu/selection_context_menu.h"
 #include "schrodinger/sketcher/model/mol_model.h"
 #include "schrodinger/sketcher/model/non_molecular_object.h"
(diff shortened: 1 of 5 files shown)


src/schrodinger/sketcher/sketcher_widget.cpp
```diff
@@ -918,7 +929,16 @@ void SketcherWidget::showContextMenu(
...
+        bool all_monomeric =
+            !filtered_atoms.empty() &&
+            std::ranges::all_of(filtered_atoms, [](const auto* atom) {
+                return is_atom_monomeric(atom);
+            });
+        if (all_monomeric) {
+            menu = m_monomer_context_menu;
+        } else {
+            menu = m_atom_context_menu;
+        }
```
ethan-schrodinger, line 941: Is this the correct behavior? Specifically: if a selection contains some monomers and some atoms, will all of the actions of the atomistic context menu apply to the monomers in a way that makes sense, or should we have a restricted hybrid context menu?

ethan-schrodinger (approved):

src/schrodinger/sketcher/sketcher_widget.cpp
JarrettSJohnson: I didn't see this behavior outlined in the spec and assumed hybrid selection likely wasn't possible if we had a monomer mode vs atom mode. But during standup today it seems like we may be leaning toward having the hybrid case so it probably needs to be defined eventually. I can submit a ticket regarding this.

KevKeating (approved):

PR #315, opened 2026-04-27
SKETCH-2648: Add Set L/D-form and Mutate monomer in context menu

JarrettSJohnson:
* Linked Case: SKETCH-2648

### Description
Adds the ability to Set L-form and D-form for that monomer(s) in context menu.

Not sure if there's a canonical way to do this(!!!), but settled on defining a structural set of rules for what classifies as a d-form monomer:
1. symbol begins with `d`
2. symbol is a `PEPTIDE`
3. symbol stripped of prefix `d` also exists in the DB as a peptide
4. both entries share the same `NATURAL_ANALOG` fields and are non-empty.

### Testing Done
Adds various tests for submenu/action visibility and labels depending on the selected aa monomer(s). Also included some tests against the d-amino acids ruleset.

### DIFF
--- a/src/schrodinger/sketcher/menu/monomer_context_menu.cpp
+++ b/src/schrodinger/sketcher/menu/monomer_context_menu.cpp
@@ -20,10 +20,7 @@ namespace schrodinger
 namespace sketcher
 {
 
-namespace
-{
-
-bool all_peptide(const std::unordered_set<const RDKit::Atom*>& atoms)
+static bool all_peptide(const std::unordered_set<const RDKit::Atom*>& atoms)
 {
     if (atoms.empty()) {
         return false;
@@ -33,7 +30,8 @@ bool all_peptide(const std::unordered_set<const RDKit::Atom*>& atoms)
     });
 }
 
-rdkit_extensions::opt_string_t peptide_natural_analog(std::string_view sym)
+static rdkit_extensions::opt_string_t
+get_peptide_natural_analog(std::string_view sym)
 {
     return rdkit_extensions::MonomerDatabase::instance().getNaturalAnalog(
         std::string(sym), rdkit_extensions::ChainType::PEPTIDE);
@@ -42,18 +40,18 @@ rdkit_extensions::opt_string_t peptide_natural_analog(std::string_view sym)
 // `dFoo` is the D-form of `Foo` iff both exist as PEPTIDEs and share
 // the same NATURAL_ANALOG. Prefix alone is unsafe — a custom DB could
 // name an entry `dXyz` with no semantic link to `Xyz`.
-bool is_d_form(std::string_view sym)
+static bool is_d_form(std::string_view sym)
 {
     if (sym.size() < 2 || sym.front() != 'd') {
         return false;
     }
-    auto sym_analog = peptide_natural_analog(sym);
-    auto stripped_analog = peptide_natural_analog(sym.substr(1));
+    auto sym_analog = get_peptide_natural_analog(sym);
+    auto stripped_analog = get_peptide_natural_analog(sym.substr(1));
     return sym_analog && stripped_analog && !sym_analog->empty() &&
            *sym_analog == *stripped_analog;
 }
 
-std::optional<std::string> toggle_d_form_symbol(std::string_view sym)
+static std::optional<std::string> toggle_d_form_symbol(std::string_view sym)
 {
     if (is_d_form(sym)) {
         // Stripping the prefix is safe: is_d_form only succeeds when
@@ -63,8 +61,8 @@ std::optional<std::string> toggle_d_form_symbol(std::string_view sym)
     // L → D: candidate "d" + sym is a true D-form when both sides
     // exist in the DB AND share the same NATURAL_ANALOG.
     auto candidate = std::string("d") + std::string(sym);
-    auto candidate_analog = peptide_natural_analog(candidate);
-    auto sym_analog = peptide_natural_analog(sym);
+    auto candidate_analog = get_peptide_natural_analog(candidate);
+    auto sym_analog = get_peptide_natural_analog(sym);
     if (candidate_analog && sym_analog && !candidate_analog->empty() &&
         *candidate_analog == *sym_analog) {
         return candidate;
@@ -72,10 +70,14 @@ std::optional<std::string> toggle_d_form_symbol(std::string_view sym)
     return std::nullopt;
 }
 
-// Target D-Form unless every atom is already D-form.
-bool should_target_d_form(const std::unordered_set<const RDKit::Atom*>& atoms)
+// True only when every atom in `atoms` is already a D-form residue.
+// Used to drive the Set D-/L-Form action label and target: when this
+// returns true, the action reads "Set L-Form" and clicking it flips
+// the residues back to L; otherwise it reads "Set D-Form".
+static bool
+all_residues_are_d_form(const std::unordered_set<const RDKit::Atom*>& atoms)
 {
-    return !std::ranges::all_of(atoms, [](const auto* a) {
+    return std::ranges::all_of(atoms, [](const auto* a) {
         return is_d_form(get_monomer_res_name(a));
     });
 }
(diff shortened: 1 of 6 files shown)


src/schrodinger/sketcher/menu/monomer_context_menu.cpp
```diff
@@ -1,17 +1,246 @@
...
+{
+    if (atoms.empty()) {
+        return false;
+    }
+    return std::ranges::all_of(atoms, [](const auto* a) {
+        return get_monomer_type(a) == MonomerType::PEPTIDE;
+    });
+}
+
+rdkit_extensions::opt_string_t peptide_natural_analog(std::string_view sym)
```
ethan-schrodinger: Nit: `peptide_natural_analog` -> `get_peptide_natural_analog`?

src/schrodinger/sketcher/menu/monomer_context_menu.cpp
```diff
@@ -1,17 +1,246 @@
...
+    return std::nullopt;
+}
+
+// Target D-Form unless every atom is already D-form.
+bool should_target_d_form(const std::unordered_set<const RDKit::Atom*>& atoms)
+{
+    return !std::ranges::all_of(atoms, [](const auto* a) {
+        return is_d_form(get_monomer_res_name(a));
+    });
+}
```
ethan-schrodinger: This function's name and behavior are not intuitive to me, maybe because it's not obvious under what conditions D-form should be targeted (and for what purpose).

What if we renamed this to `all_residues_are_d_form()` and inverted the expected output? In that case, we'd get the same functionality but the expected behavior would be totally clear.

src/schrodinger/sketcher/menu/monomer_context_menu.cpp
```diff
@@ -1,17 +1,246 @@
...
+void MonomerContextMenu::createSetDFormAction()
+{
+    m_set_d_form_action = addAction("Set D-Form", this, [this]() {
+        const bool target_d_form = should_target_d_form(m_atoms);
+        auto batch = build_d_form_batch(m_atoms, target_d_form);
+        if (batch.empty()) {
+            return;
+        }
+        emit mutateResidueRequested(
+            std::move(batch), target_d_form ? "Set D-Form" : "Set L-Form");
```
ethan-schrodinger, line 202: Under what conditions would we expect a selection to contain a mixture of D-form and L-form? Specifically, is it normal to have a structure that contains a mixture of the two?

cdvonbargen:
@i-tub @rachelnwalker -- is there a canonical way to classify d-form monomers? I swear I heard in passing Ivan was doing something here with atomistic/monomeric interconversion stuff

JarrettSJohnson:
If there is, please let me know! This was the most involved part and would cut this PR by a lot if so.

src/schrodinger/sketcher/menu/monomer_context_menu.cpp
JarrettSJohnson: Probably more of a design question, but can I flip the question and ask why we wouldn't? I imagine that you could design a peptide where one site is D-form because you've found that it improved binding or prevents higher-order structuring, etc...

src/schrodinger/sketcher/menu/monomer_context_menu.cpp
JarrettSJohnson: I battled with both of these ideas, and asked Claude which preferred `should_target_d_form` since at the call-site the question that we're asking is "Do we satisfy the conditions where we should show the user 'Set D-form'?", then the body of this function becomes just an implementation detail (and could be changed later without renaming the function). I'm indifferent either way and certainly happy to change it if it makes the intent clearer.

rachelnwalker:
> @i-tub @rachelnwalker -- is there a canonical way to classify d-form monomers? I swear I heard in passing Ivan was doing something here with atomistic/monomeric interconversion stuff

I think that @JarrettSJohnson 's rules make sense here, but this will definitely only work for out built-in monomers. There is no restriction or convention on the naming schemes for user created monomers.

src/schrodinger/sketcher/menu/monomer_context_menu.cpp
KevKeating: I have a very slight preference for `all_residues_are_d_form()`, but I'm fine with either option if anyone else feels strongly one way or the other.

src/schrodinger/sketcher/menu/monomer_context_menu.cpp
```diff
@@ -1,17 +1,246 @@
...
+#include "schrodinger/sketcher/model/sketcher_model.h"
+#include "schrodinger/sketcher/rdkit/monomeric.h"
+
 namespace schrodinger
 {
 namespace sketcher
 {
 
+namespace
+{
```
KevKeating: The [core suite style guide](https://schrodinger.atlassian.net/wiki/spaces/SOFTWAREDEV/pages/48138831/Schr+dinger+C+and+C+Style+Guide#Prefer-static-function-over-unnamed-namespaces) says to prefer static functions over anonymous namespaces (although we definitely already have a handful of anonymous namespaces in Sketcher code).

src/schrodinger/sketcher/sketcher_widget.cpp
```diff
@@ -66,6 +66,41 @@ namespace schrodinger
...
+};
+
+/**
+ * Capture atom identity by index before a sequence of
+ * `MolModel::mutateMonomers` calls. Indices are stable since the
+ * underlying `mutateMonomer` only edits properties (no atom
+ * add/remove/reorder). The caller resolves each batch's indices back
+ * to live atoms via `getMol()->getAtomWithIdx(...)` immediately
+ * before each mutation.
+ */
```
KevKeating, line 93: You could mention that this call will invalidate the passed-in vector

src/schrodinger/sketcher/sketcher_widget.cpp
```diff
@@ -745,6 +780,38 @@ void SketcherWidget::connectContextMenu(const MonomerContextMenu& menu)
...
+                if (mutations.empty()) {
+                    return;
+                }
+                auto indexed = capture_atom_indices(std::move(mutations));
+
+                // Multi-pair batches (the D-/L-Form toggle when residues map to
+                // multiple target symbols) get wrapped in a single QUndoStack
+                // macro so the user sees one undo entry.
+                const bool batch = indexed.size() > 1;
+                if (batch) {
```
KevKeating: A macro with only one subcommand is still valid, so you could just create a macro here even if size == 1 if you wanted.  Also, MolModel has a `createUndoMacro` method that returns an RAII object for the macro so you don't need to worry about manually ending it.

There's nothing wrong with the current code, though, if you prefer that.

KevKeating (approved):
I just had a few minor nits.

JarrettSJohnson:
> > @i-tub @rachelnwalker -- is there a canonical way to classify d-form monomers? I swear I heard in passing Ivan was doing something here with atomistic/monomeric interconversion stuff
> 
> I think that @JarrettSJohnson 's rules make sense here, but this will definitely only work for out built-in monomers. There is no restriction or convention on the naming schemes for user created monomers.

This is also likely going to force me to come up with a set of heuristics for protonation, methylation, etc...as well.

Should I just come up with these rules (based off of the default DB) as I continue with the context menus or file a separate ticket for these to be figured out first?

Related question, are we able to change the DB schema to support this explicitly for custom and non-custom analogs?

src/schrodinger/sketcher/menu/monomer_context_menu.cpp
ethan-schrodinger: Either is fine with me as long as the intent is made clear in the docs somehow, including what it means to be "targeted" and under what conditions that's appropriate.

ethan-schrodinger (approved):
LGTM, pending more info from @rachelnwalker / @i-tub.

rachelnwalker:
> 
> This is also likely going to force me to come up with a set of heuristics for protonation, methylation, etc...as well.

Interesting - is this an ask from the sketcher UI? I didn't know that was a part of the spec, I would say to file a SHARED ticket. I assume we'd want to precalculate these fields on DB load

JarrettSJohnson:
> > This is also likely going to force me to come up with a set of heuristics for protonation, methylation, etc...as well.
> 
> Interesting - is this an ask from the sketcher UI? I didn't know that was a part of the spec, I would say to file a SHARED ticket. I assume we'd want to precalculate these fields on DB load

image

I'll file a ticket.

PR #317, opened 2026-04-28
SKETCH-2618: Refactoring DrawMonomerSceneTool to create AbstractMonomerSceneTool

KevKeating:
* Linked Case: SKETCH-2618

# This PR only contains moved code.  There's no new  code here.

### Description
As part of creating the scene tool for drawing full nucleotides, I refactored DrawMonomerSceneTool so that it could share a bunch of code with the new nucleotide scene tool.  To make things a bit easier to read, I've split the changes up into two separate PRs: this PR contains only the refactoring, and all of the new code will be part of a separate PR.  I also did a bit of refactoring to MonomerHintFragmentItem, as well as the unit tests.

### Testing Done
Confirmed that all of the tests still pass.

PR #318, opened 2026-04-28
SKETCH-2618: Implementing DrawMonomerFragmentSceneTool and hooking up full nucleotide tools

KevKeating:
* Linked Case: SKETCH-2618

### Description
This PR implements a DrawMonomerFragmentSceneTool and uses that to hook up the full nucleotide tools.  Note that this PR builds on the refactoring done in #317.  The scene tool class itself doesn't contain any nucleotide-specific logic, so we'll be able to use it for a generic monomeric fragment library if/when that's implemented.  The only information the class needs is:
- an RDKit mol of the fragment
- a data structure describing how the fragment connects to existing structure (i.e. which attachment point of a what fragment atom index binds to which attachment point of what monomer type)
- when clicking on empty space, which fragment atom index should be placed at the actual click coordinates (which is the sugar monomer for nucleotides)
- when clicking-and-dragging on empty space, what angle in the fragment structure should follow the cursor.  (For nucleotides, this is set so that the base follows the cursor.)

The only part of the nucleotide scene tool that's not implemented here is the cursor hint.  There's logic for when to show it and hide it (since we want to hide it when a blue hint structure fragment is shown in the workspace), but no actual logic to generate it in the first place.  I'll work on that next.

It also might be worth making the side bar buttons look a little nicer.  Right now they use a HELM string, although it's not actually the correct HELM string; the buttons show the sugar and base of one nucleotide with the phosphate from the _next_ nucleotide instead of the phosphate from that nucleotide.  We could change the HELM string to something like `P.R(A)` instead of `R(A)P, which would be accurate, although it would probably be more recognizable if we drew shapes around the letters, something like
$\huge \textⓅ-🅁-◇$ (except not using hacky Unicode so it actually looks right and contains the base inside of the diamond).  I can file a spin off case for that, since it doesn't seem like it needs to be an exceedingly high priority.

Here's a screenshot of what the scene tool looks like when adding on to an existing nucleotide:
image

### Testing Done
Tested via the desktop app on Windows and added new integration-level tests.


src/schrodinger/sketcher/molviewer/coord_utils.h
```diff
@@ -216,10 +216,12 @@ SKETCHER_API bool are_points_on_same_side_of_line(const QPointF& point1,
...
 /**
  * @return the angle in radians of a line from start to end, rounded to the
- * nearest pi/6 (30 degree) increment.
+ * nearest pi/num_increments increment. The default num_increments value
+ * corresponds to 30 degree increments.
  */
 SKETCHER_API qreal get_rounded_angle_radians(const QPointF& start,
-                                             const QPointF& end);
+                                             const QPointF& end,
+                                             const int num_increments = 12);
```
ethan-schrodinger, line 224: Nit: is there a reason to think that we'll continue to rely on this default in the future? If not, it may be better to not specify a default value, and just update the current calls to this function so that they explicitly pass `num_increments=12`.

ethan-schrodinger (approved):
>it would probably be more recognizable if we drew shapes around the letters, something like
Ⓟ−🅁−◇
 (except not using hacky Unicode so it actually looks right and contains the base inside of the diamond). I can file a spin off case for that, since it doesn't seem like it needs to be an exceedingly high priority.

+1. This all sounds right to me.

src/schrodinger/sketcher/molviewer/coord_utils.h
KevKeating: I think any calls from atomistic code will continue to use 12.  It's only calls from monomeric code that would want anything different.

PR #319, opened 2026-04-30
Add a BUILD.md

i-tub:
This has the procedure that I found to work for building on Linux. If there's a better way, I'm all ears!

Also, I suppose we could put this in a script, Makefile, or something, and that we could include a Dockerfile, but I'm not sure how far we want to go with this. For now, these are the instructions I wish I had.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -34,3 +34,5 @@ TAGS
 # compilation database
 compile_commands.json
 .vscode/
+.venv
+ext_bld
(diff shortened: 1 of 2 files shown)


cdvonbargen (approved):

cdvonbargen (approved):
Works for me. My only concern is this drifts from the build instructions in the GH workflows, but agreed that's a problem we can solve another day

PR #320, opened 2026-04-30
SKETCH-2498: Add release workflow

cdvonbargen:
* Linked Case: SKETCH-2498

### Description

From our conversation today -- adds a GH action for creating a release tag and uploading the macos/linux/windows/wasm assets to https://github.com/schrodinger/sketcher/releases. If we need to patch, it's possible to rerun the workflow -- if a release for the version already exists, the tag is moved to the current branch HEAD and the artifacts are updated in place.

### Testing Done

I can't run this until it's committed....

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -1,0 +1,53 @@
+name: Release
+
+on:
+  workflow_dispatch:
+    inputs:
+      version:
+        description: 'Release version corresponding to the branch to use'
+        required: true
+        type: string
+
+jobs:
+  build:
+    uses: ./.github/workflows/sketcher-builder.yml
+    with:
+      ref: ${{ inputs.version }}
+
+  release:
+    needs: build
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@v4
+        with:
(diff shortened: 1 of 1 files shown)


i-tub (approved):
Looks plausible; let's see what happens when it runs!

.github/workflows/release.yml
```diff
@@ -0,0 +1,50 @@
...
+    with:
+      ref: ${{ inputs.version }}
+
+  release:
+    needs: build
+    runs-on: ubuntu-latest
+    permissions:
+      contents: write
+    steps:
+      - uses: actions/checkout@v4
```
KevKeating: This should use a newer version of checkout (probably v6) to avoid the Node.js 20 warnings.

.github/workflows/release.yml
```diff
@@ -0,0 +1,50 @@
...
+        with:
+          ref: ${{ inputs.version }}
+
+      - name: Download release assets
+        uses: actions/download-artifact@v4
```
KevKeating: You should use an up-to-date version of this action as well.

KevKeating (approved):

PR #323, opened 2026-04-30
SKETCH-2729: Add nucleic acid analogs to the monomer palette

JarrettSJohnson:
* Linked Case: SKETCH-2729

### Description
Similar to #279 , but adds support for nucleic acid analogs. A couple of unique logic needed for this PR:

1. Unlike the amino acids, nucleic acid analogs are generally absent from the DB. For consistency, I also converted each nucleic acid button to a `ModularPopupButton` to a) support custom na analogs b) tooltips are populated automatically (which in turns also solves the issue in SKETCH-2689).
2. Updating the buttons when switching analogs needs special logic for ensuring that analogs that exist in the toolbar are not affected by other buttons that could share the same analog. Example: Since we have deoxyribose `dR` as a top-level button in the toolbar (rather than having it under `R`), selecting `dR` would inadvertently also convert the `R` button to switch to its `dR` analog -- which I assume is not the expected behavior.

### Testing Done
Tests various aspects like analog mutation and popup population.

### DIFF
--- a/src/schrodinger/sketcher/widget/monomer_tool_widget.cpp
+++ b/src/schrodinger/sketcher/widget/monomer_tool_widget.cpp
@@ -27,6 +27,48 @@ namespace sketcher
 using AnalogMap =
     std::unordered_map<std::string, std::vector<rdkit_extensions::MonomerInfo>>;
 
+// dR has NATURAL_ANALOG=R, so it's not a key in the analog map. Pull
+// R's analogs (and R itself) into dR's popup for symmetry with R, which
+// already shows dR.
+static const std::unordered_map<std::string, std::string>
+    NA_ANALOG_LOOKUP_OVERRIDES = {{"dR", "R"}};
+
+/**
+ * Build the analog list for an NA button: optionally prepend a parent
+ * standard (e.g. R for the dR button) sourced from `standard_names`,
+ * then append the natural-analog list keyed by the override target (or
+ * `symbol` itself if no override applies), filtering out entries whose
+ * symbol matches `symbol` (the button's own standard).
+ */
+static std::vector<rdkit_extensions::MonomerInfo> get_analogs_for_na_button(
+    const std::string& symbol, const AnalogMap& analogs_by_na,
+    const std::unordered_map<std::string, std::string>& standard_names)
+{
+    std::vector<rdkit_extensions::MonomerInfo> analogs;
+    auto override_it = NA_ANALOG_LOOKUP_OVERRIDES.find(symbol);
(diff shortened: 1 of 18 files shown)


src/schrodinger/sketcher/widget/monomer_tool_widget.cpp
```diff
@@ -265,6 +344,46 @@ void MonomerToolWidget::updateCheckedButton()
         }
     }
 
+    // Show the popup indicator only on the currently checked NA button
+    for (auto& [btn, popup] : m_nucleic_acid_symbol_popups) {
+        auto* modular_btn = qobject_cast<ModularToolButton*>(btn);
+        Q_ASSERT_X(modular_btn, "MonomerToolWidget",
+                   "Expected ModularToolButton");
```
ethan-schrodinger: Should we keep these asserts in, or were they only meant to be here during development?

ethan-schrodinger (commented):
>Since we have deoxyribose dR as a top-level button in the toolbar (rather than having it under R), selecting dR would inadvertently also convert the R button to switch to its dR analog -- which I assume is not the expected behavior.

I think your assumption is correct, but it also sort of points to a sharp edge in the current spec. Do we think it make sense to have both R and dR on the primary palette?

src/schrodinger/sketcher/widget/monomer_tool_widget.cpp
JarrettSJohnson: From my understanding, they're like regular asserts that are stripped out of release builds, but I seem to be the only one who has been using `Q_ASSERTS`; so I don't mind removing them for consistency.

JarrettSJohnson:
> > Since we have deoxyribose dR as a top-level button in the toolbar (rather than having it under R), selecting dR would inadvertently also convert the R button to switch to its dR analog -- which I assume is not the expected behavior.
> 
> I think your assumption is correct, but it also sort of points to a sharp edge in the current spec. Do we think it make sense to have both R and dR on the primary palette?

IMO, I think it's smoother UX to keep both in.

cdvonbargen (approved):

KevKeating (commented):
With the current change, does the dR button get any analogs?  If not, I think we probably want to give it all of the same analogs as R.  Would that break anything?

JarrettSJohnson:
> With the current change, does the dR button get any analogs? If not, I think we probably want to give it all of the same analogs as R. Would that break anything?

added

JarrettSJohnson:
Woops, the re-review request for @cdvonbargen was accident (was already approved)

ethan-schrodinger (approved):

cdvonbargen (approved):

KevKeating (approved):

PR #325, opened 2026-05-01
SKETCH-2716: Prevent cleanup for monomeric structures

JarrettSJohnson:
* Linked Case: SKETCH-2716

### Description
As predicted by @KevKeating , this just needed monomer property flag to prevent monomer coordinates from changing during cleanup.

### Testing Done
Added test to ensure cleanup doesn't affect monomer coordinates.

### DIFF
--- a/src/schrodinger/sketcher/model/mol_model.cpp
+++ b/src/schrodinger/sketcher/model/mol_model.cpp
@@ -2932,6 +2932,9 @@ void MolModel::addMolCommandFunc(RDKit::ROMol mol)
         add_empty_3d_conformer(mol);
     }
     m_mol.insertMol(mol);
+    if (is_monomeric) {
+        m_mol.setProp(HELM_MODEL, true);
+    }
 
     for (auto& sgroup : getSubstanceGroups(m_mol)) {
         setTagForSGroup(sgroup, m_next_s_group_tag++);
--- a/src/schrodinger/sketcher/model/mol_model.cpp
+++ b/src/schrodinger/sketcher/model/mol_model.cpp
@@ -1478,6 +1478,24 @@ static void update_post_compute2DCoords(RDKit::RWMol& mol)
     fix_variable_attachment_bond_coordinates(mol);
 }
 
+/**
+ * Sync the mol-level HELM_MODEL property on m_mol to match its current
+ * monomeric state before a coord regeneration. MolModel tracks monomeric
+ * state per-atom (SKETCHER_ATOM_IS_MONOMERIC);
+ * rdkit_extensions::compute2DCoords keys off the mol-level HELM_MODEL property
+ * to choose between the monomer and atomistic layout, so we sync them here so
+ * the correct algorithm runs regardless of how the model was built up or torn
+ * down (SKETCH-2716).
+ */
+static void sync_helm_model_property(bool is_monomeric, RDKit::RWMol& mol)
+{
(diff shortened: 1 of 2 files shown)


ethan-schrodinger (approved):

src/schrodinger/sketcher/model/mol_model.cpp
```diff
@@ -2932,6 +2932,9 @@ void MolModel::addMolCommandFunc(RDKit::ROMol mol)
         add_empty_3d_conformer(mol);
     }
     m_mol.insertMol(mol);
+    if (is_monomeric) {
+        m_mol.setProp(HELM_MODEL, true);
+    }
```
KevKeating: This won't set the property if the user builds a monomeric molecule by adding a single monomer at a time via `addMonomer`.  I think the better option would be to set the `HELM_MODEL` in `MolModel::cleanUpSelection` and `MolModel::regenerateCoordinates` and then clear it at the end of those methods.  MolModel itself intentionally doesn't expect `HELM_MODEL` to be set on monomeric models since it uses it's own per-atom `SKETCHER_ATOM_IS_MONOMERIC` property (since HELM_MODEL can only be set for the whole structure and MolModel allows mixing monomeric and atomistic atoms, although the interface currently imposes the one-or-the-other requirement).  As a result, I think it's safer to set `HELM_MODEL` before passing the molecule to the outside world rather than trying to keep HELM_MODEL up-to-date on `m_mol`.  That's how `getMolForExport` handles things currently.

JarrettSJohnson: > This won't set the property if the user builds a monomeric molecule by adding a single monomer at a time via addMonomer

It seems like it does indirectly via `addMonomer` -> `assignChains` -> `monomer_mol.setProp("HELM_MODEL", true);`.

https://github.com/schrodinger/sketcher/blob/ee3307d64c968f29955999e04614ea66fb86e419/src/schrodinger/sketcher/model/mol_model.cpp#L680

https://github.com/schrodinger/sketcher/blob/ee3307d64c968f29955999e04614ea66fb86e419/src/schrodinger/rdkit_extensions/monomer_mol.cpp#L636

JarrettSJohnson: I could perhaps enforce this works with `addMonomers` with a test and/or also add your suggestion of setting it explicitly in `MolModel::cleanUpSelection` and `MolModel::regenerateCoordinates`

KevKeating: > It seems like it does indirectly via addMonomer -> assignChains -> monomer_mol.setProp("HELM_MODEL", true);

Ahh, that's a good point.  I hadn't realize that `assignChains` would set that.  I think I'd still lean towards adding something to `MolModel::cleanUpSelection` and `MolModel::regenerateCoordinates`, since ensuring that the property is consistently in the correct state in MolModel seems like it could be tricky.  For example, if someone pasted a HELM string, then erased all of the monomers, and then drew an atomistic model, we presumably need to make sure that HELM_MODEL is cleared before trying to recalculate coordinates.  (We should probably also make sure that `getMolForExport` clears the HELM_MODEL property as needed now that you've pointed out it can get set in `assignChains`.)

JarrettSJohnson: done.

KevKeating (approved):

PR #328, opened 2026-05-01
SKETCH-2618: Adding cursor hint for nucleotide tools

KevKeating:
* Linked Case: SKETCH-2618

### Description
The nucleotide tools now have a cursor hint.  The cursor hint is generated from the RDKit mol, so it'll work for any monomeric fragments if/when this tool gets hooked up to a fragment library.  I made the cursor hint slightly larger than a typical cursor hint so that the labels would be readable-ish.  Also, because the nucleotide structure is L-shaped, it doesn't fill up the whole square of the pixmap, so it winds up with about the same visual "weight" as a normal cursor hint despite the larger size.  Here's what the hint looks like on Windows:

image

### Testing Done
Tested on the Windows desktop app.


src/schrodinger/sketcher/molviewer/scene_utils.h
```diff
@@ -138,9 +138,13 @@ SKETCHER_API QPixmap cursor_hint_from_svg(const QString& path,
  * graphics_item is small, this will effectively zoom out when generating the
  * cursor hint. If the bounding rect of graphics_item is larger than
  * min_scene_size, then this setting will have no effect.
+ * @param cursor_hint_scale_size A scale to apply to the size of the cursor hint
```
JarrettSJohnson: Nit: `scale_size` in docstring but `size_scale` in param name.

JarrettSJohnson (approved):
Cool!

ethan-schrodinger (approved):

src/schrodinger/sketcher/molviewer/scene_utils.h
KevKeating: Fixed

PR #331, opened 2026-05-04
SKETCH-2497: Add bump-quarter mode to bump-version action

cdvonbargen:
* Linked Case: SKETCH-2497

### Description

@brandon-wick wired up cutting the sketcher branch for me, but I realized that we still needed a way to bump/reset the 20YY.Q.BBB on version.json

This adds a arg to the bump-version action that, when called with `main`, will roll the quarter up. Brandon should be able to add a call to this action on the sketcher repo on branch cut similar to how Lenna did for the NBs:
https://github.com/schrodinger/core-suite-ci/blob/c51b6b9ef9ffa9dbd2e065edc6584dda32f74c35/.github/workflows/merge_trigger.yml#L155
```
      - name: Reset the version on main branch
        uses: ./.github/actions/bump-version
        with:
          branch: main
          bump-quarter: true
          dry-run: ${{ inputs.dry_run }}
```

### Testing Done

Added a test, and wired github_actions tests up to pre-commit (which seemed like the simplest thing to do)


.github/actions/bump-version/bump_version.py
```diff
@@ -0,0 +1,57 @@
...
+def next_version(current: str, bump_quarter: bool = False) -> str:
+    """Return the version that should follow ``current``.
+
+    ``current`` is expected in ``YYYY.Q.B`` form.
+    Raises ``ValueError`` on malformed input or out-of-range quarter.
+    """
+    parts = current.split(".")
+    if len(parts) != 3:
+        raise ValueError(
+            f"expected version in YYYY.Q.B form, got {current!r}")
```
KevKeating: I'm surprised by the `!r` formatting on `current`.  That will print the repr() of a string instead of just printing the string itself.  I'm assuming this was something Claude added?  Maybe it wanted the quotes to appear in the output, which repr() would do?

I don't have any objections to it.  Just unexpected.

test/github_actions/test_bump_version.py
```diff
@@ -0,0 +1,62 @@
...
+import sys
+
+import pytest
+
+SKETCHER_REPO_ROOT = pathlib.Path(__file__).resolve().parents[2]
+BUMP_VERSION_DIR = (SKETCHER_REPO_ROOT / ".github" / "actions" /
+                    "bump-version")
+sys.path.insert(0, str(BUMP_VERSION_DIR))
+import bump_version  # noqa: E402
+
```
KevKeating, line 14: You might want to add a `sys.path.pop(0)` here so that `sys.path` gets returned to normal after the import.  It would only really matter if there were other files being read here, which there aren't, but it's probably worth it in case we add additional tests later.

KevKeating (approved):

brandon-wick (approved):
LGTM

nit: pin pytest to match mmshare (`pytest==9.0.2`) on line 27 of sketcher-builder.yml to avoid surprise breakage from an unpinned dependency.

test/github_actions/test_bump_version.py
cdvonbargen: Good call — done in 17ae1301. Wrapped the import in try/finally to pop `sys.path` back to normal.

.github/actions/bump-version/bump_version.py
cdvonbargen: Yeah, that was Claude. The intent was to make whitespace-only or empty inputs visible in the error message (`got ''` rather than `got`). Happy to drop the `!r` if you'd prefer plain string formatting — let me know.

PR #334, opened 2026-05-07
SKETCH-2750: Refresh HELM_MODEL stale state when dealing with atoms

JarrettSJohnson:
- Linked Case: SKETCH-2750

### Description
Bug occurred when attempting to cut/paste atoms after being in monomer mode for a while: 'Atom has no monomer info'.

Tracked it down to `HELM_MODEL` state not being properly cleared--which is now done in `strip_notes_and_mol_model_tags`. Monomer state is checked again so that it isn't stale over time.

### Testing Done
Added a test that reproduced the issue from my steps programmatically.
```
    BOOST_TEST(!contains_monomeric_atom(*exported));
    BOOST_TEST(!rdkit_extensions::isMonomeric(*exported));
    BOOST_TEST(!exported->hasProp(HELM_MODEL));
```
Should now be correct for atomistic mode after dealing with monomers.

### DIFF
--- a/src/schrodinger/sketcher/model/mol_model.cpp
+++ b/src/schrodinger/sketcher/model/mol_model.cpp
@@ -161,6 +161,7 @@ void strip_notes_and_mol_model_tags(RDKit::ROMol& mol)
         bond->clearProp(RDKit::common_properties::bondNote);
     }
     mol.clearProp(RDKit::common_properties::molNote);
+    mol.clearProp(HELM_MODEL);
     mol.clearAllAtomBookmarks();
     mol.clearAllBondBookmarks();
 
@@ -225,11 +226,14 @@ boost::shared_ptr<RDKit::ROMol> MolModel::getMolForExport() const
 {
     // TODO: add API to export selection as atom/bond properties
     auto mol_copy = boost::make_shared<RDKit::ROMol>(m_mol);
-    if (contains_monomeric_atom(m_mol)) {
+    // Determine HELM-ness from the selection's atom content before stripping,
+    // since stripping also clears the per-atom monomeric flags.
+    const bool is_monomeric = contains_monomeric_atom(*mol_copy);
+    strip_notes_and_mol_model_tags(*mol_copy);
+    if (is_monomeric) {
         // RDKit uses a mol-level property to indicate HELM mols
         mol_copy->setProp(HELM_MODEL, true);
     }
-    strip_notes_and_mol_model_tags(*mol_copy);
     return mol_copy;
 }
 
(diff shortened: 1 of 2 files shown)


src/schrodinger/sketcher/model/mol_model.cpp
```diff
@@ -225,11 +226,14 @@ boost::shared_ptr<RDKit::ROMol> MolModel::getMolForExport() const
 {
     // TODO: add API to export selection as atom/bond properties
     auto mol_copy = boost::make_shared<RDKit::ROMol>(m_mol);
-    if (contains_monomeric_atom(m_mol)) {
+    // Determine HELM-ness from the selection's atom content before stripping,
+    // since stripping also clears the per-atom monomeric flags.
+    const bool is_monomeric = contains_monomeric_atom(*mol_copy);
```
ethan-schrodinger, line 231: >The molecule-level HELM_MODEL property gets set when a monomer is added but is never cleared when the monomer is deleted.
 
Maybe a dumb question, but if we can just run `contains_monomeric_atom()` whenever we need to determine whether the sketcher contents are monomeric, what is the purpose of applying this `HELM_MODEL` property in the sketcher, rather than just relying on `contains_monomeric_atom()` to determine the correct value on the fly?

ethan-schrodinger (approved):

src/schrodinger/sketcher/model/mol_model.cpp
JarrettSJohnson: From my brief look-through (perhaps Rachel or Kevin has more insight here), is that `contains_monomeric_atom` is a per-atom sketcher-internal flag to express if it's monomeric whereas `HELM_MODEL` is a `mol`-level property flag used through RDKit to enforce monomeric behavior. Incoming fragments can be monomeric but the nascent `Mol` at that time (before attachment) may not yet be so they're not quite interchange-able from what I see.

KevKeating: Yeah, Jarrett is exactly right.  The `HELM_MODEL` property is what `rdkit_extensions` uses to distinguish between atomistic and monomeric models.  There's currently no "real" way to set atomistic versus monomeric on a per-atom basis, so `MolModel` uses it's own `SKETCHER_ATOM_IS_MONOMERIC` property.  Since nothing else knows about that property, though, we strip it and convert to `HELM_MODEL` on export.

KevKeating (commented):
We should probably also clear the `HELM_MODEL` property on `m_mol` after any `assignChains(m_mol)` calls, since I think that's how it's getting set in the first place.  That would help to avoid any bugs where something only works if the model happens to have been passed to `assignChains` first.  (It's not a bad idea to also clear the property in `strip_notes_and_mol_model_tags`, just in case.)

KevKeating (approved):

PR #341, opened 2026-05-09
SKETCH-2751: Support monomeric cut/copy/paste

JarrettSJohnson:
- Linked Case: SKETCH-2751

### Description
Supports copy and paste of monomers. Perhaps the main expected difference is that I'm using `RDMOL_BINARY_BASE64` rather than `HELM` so that the relative positions of each monomer is maintained.

### Testing Done
Tests partial and full cut and paste as well as partial and full copy and paste


src/schrodinger/sketcher/menu/cut_copy_action_manager.cpp
```diff
@@ -63,6 +63,17 @@ SceneSubset CutCopyActionManager::getSubset()
     }
 }
 
+Format CutCopyActionManager::getCutCopyFormat() const
+{
+    if (m_sketcher_model->getMoleculeType() == MoleculeType::MONOMERIC) {
+        // Pickle round-trip preserves the monomeric structure and the 2D
+        // conformer, so pasted monomers land at the same positions instead
+        // of being re-laid-out by the chain layout engine.
+        return Format::RDMOL_BINARY_BASE64;
```
KevKeating: Chris has previously mentioned that we don't want to expose this format to users.  @cdvonbargen, I'm assuming that's still the case?  Should we just default to HELM strings here?

A alternate possibility would be to tweak `SketcherWidget::copy`/`setClipboardContents` so that when we copy a HELM string (and possibly SMILES strings as well?), we _also_ add RDMOL_BINARY_BASE64 data to the clipboard with a separate MIME type.  Then `SketcherWidget::getClipboardContents` would look for that MIME type first and return it if found.  That way, the Sketcher would receive RDMOL_BINARY_BASE64 data on paste but any other application would only see the HELM /SMILES string.

JarrettSJohnson: Def agreed. done.

src/schrodinger/rdkit_extensions/convert.cpp
```diff
@@ -788,7 +788,11 @@ std::string to_string(const RDKit::ROMol& input_mol, const Format format)
         auto is_monomeric = isMonomeric(input_mol);
         auto is_seq_format =
             std::ranges::find(SEQ_FORMATS, format) != SEQ_FORMATS.end();
-        if (is_monomeric && !is_seq_format) {
+        if (is_monomeric && !is_seq_format &&
+            format != Format::RDMOL_BINARY_BASE64) {
+            // RDMOL_BINARY_BASE64 is a lossless pickle round-trip; preserve
+            // the monomeric mol (HELM_MODEL prop and conformer) instead of
+            // downgrading to atomistic.
```
rachelnwalker, line 795: 👍🏻

rachelnwalker (approved):
Just approved the bit in convert.cpp :)

KevKeating (approved):

ethan-schrodinger (approved):

src/schrodinger/sketcher/menu/cut_copy_action_manager.cpp
```diff
@@ -63,6 +63,16 @@ SceneSubset CutCopyActionManager::getSubset()
...
 }
 
+Format CutCopyActionManager::getCutCopyFormat() const
+{
+    if (m_sketcher_model->getMoleculeType() == MoleculeType::MONOMERIC) {
+        // HELM is the user-visible payload; SketcherWidget::copy also stashes
+        // a lossless pickle for intra-sketcher pastes (see its comment).
+        return Format::HELM;
+    }
+    return DEFAULT_FORMAT;
```
cdvonbargen: should we just remove this constant since this is now effectively DEFAULT_ATOMISTIC_FORMAT? I'd rather see HELM/V3K in one spot.

Also, should the function be getDefaultCutCopyFormat

src/schrodinger/sketcher/sketcher_widget.cpp
```diff
@@ -567,13 +579,20 @@ void SketcherWidget::cut(Format format)
...
     std::string text;
+    std::string binary;
     try {
         text = extract_string(m_mol_model, m_sketcher_model, format, subset);
+        // For monomers also stash a pickle so intra-sketcher pastes preserve
+        // structure and 2D coords (chain layout otherwise regenerates them).
+        if (m_mol_model->isMonomeric()) {
+            binary = extract_string(m_mol_model, m_sketcher_model,
+                                    Format::RDMOL_BINARY_BASE64, subset);
+        }
```
cdvonbargen: I think Im ok with the intra-sketcher mime data, but
(a) don't know that it's a common usage pattern, so Im curious how much this really helps?
(b) think it's weird that we only do this for monomers -- shouldnt this be an all-or-nothing thing?

include/schrodinger/sketcher/sketcher_widget.h
```diff
@@ -387,10 +387,12 @@ class SKETCHER_API SketcherWidget : public QWidget
...
      * Methods for getting and setting the system clipboard. These methods are
      * overriden in the unit test since using the real clipboard can lead to
-     * intermittent failures on buildbot.
+     * intermittent failures on buildbot. The optional `binary` payload is
+     * stashed under a sketcher-private MIME type and preferred on paste.
      */
     virtual std::string getClipboardContents() const;
-    virtual void setClipboardContents(std::string text) const;
+    virtual void setClipboardContents(std::string text,
+                                      std::string binary = "") const;
```
cdvonbargen: Do we still need the default empty string here? Should we make `binary` required?

cdvonbargen (approved):

src/schrodinger/sketcher/menu/cut_copy_action_manager.cpp
```diff
@@ -63,6 +61,16 @@ SceneSubset CutCopyActionManager::getSubset()
     }
 }
 
+Format CutCopyActionManager::getDefaultCutCopyFormat() const
+{
+    if (m_sketcher_model->getMoleculeType() == MoleculeType::MONOMERIC) {
+        // HELM is the user-visible payload; SketcherWidget::copy also stashes
```
d-b-w, line 67: ah, OK. I was gonna ask about this in the other review. cool.

d-b-w (approved):
great

PR #343, opened 2026-05-10
SKETCH-2742: Export monomers to file (HELM, FASTA, atomic)

JarrettSJohnson:
- Linked Case: SKETCH-2742

### Description
Supports monomers exported as both HELM and FASTA. Additionally, atomic file formats are supported (monomers are autoconverted to atomitistic models first).

### Testing Done
Added tests which verifies the new file formats during monomer export. Also testing round-tripping for HELM, FASTA, and atomistic.

### DIFF
--- a/src/schrodinger/sketcher/dialog/file_import_export.cpp
+++ b/src/schrodinger/sketcher/dialog/file_import_export.cpp
@@ -74,7 +74,7 @@ FormatList<Format> get_import_formats()
 
 FormatList<Format> get_standard_export_formats()
 {
-    std::vector<std::tuple<Format, std::string>> mol_export_formats = {
+    std::vector<std::tuple<Format, std::string>> mol_and_seq_export_formats = {
         // Forbid MDL_MOLV2000 on export; potential stereo ambiguities
         {Format::MDL_MOLV3000, "MDL SD V3000"},
         {Format::MAESTRO, "Maestro"},
@@ -87,11 +87,16 @@ FormatList<Format> get_standard_export_formats()
         {Format::PDB, "PDB"},
         {Format::XYZ, "XYZ"},
         {Format::MRV, "Marvin Document"},
+        // Sequence formats are always offered; rdkit_extensions::to_string
+        // handles atomistic <-> monomeric conversion on the fly, and conversion
+        // failures surface through the dialog's existing error path.
+        {Format::HELM, "HELM"},
+        {Format::FASTA, "FASTA"},
     };
 
     FormatList<Format> export_formats;
-    for (const auto& [format, label] : mol_export_formats) {
-        auto extensions = rdkit_extensions::get_mol_extensions(format);
+    for (const auto& [format, label] : mol_and_seq_export_formats) {
+        auto extensions = rdkit_extensions::get_mol_and_seq_extensions(format);
         export_formats.push_back({format, label, extensions});
     }
     return export_formats;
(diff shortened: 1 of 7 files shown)


src/schrodinger/sketcher/model/sketcher_model.h
```diff
@@ -739,6 +744,11 @@ class SKETCHER_API SketcherModel : public QObject
      */
     unsigned int reactionCountRequested() const;
 
+    /**
+     * @return whether the associated scene contains a monomeric model
+     */
+    bool monomericQueryRequested() const;
```
KevKeating: This might be clearer as something like `isMonomericRequested`.  The word "query" here initially makes me think of query atoms, not that the signal is a query.

KevKeating (approved):
Looks good to me other than one nit.

ethan-schrodinger (approved):

src/schrodinger/sketcher/dialog/file_import_export.cpp
```diff
@@ -90,6 +90,14 @@ FormatList<Format> get_standard_export_formats()
...
 
     FormatList<Format> export_formats;
+    if (is_monomeric) {
+        export_formats.push_back(
+            {Format::HELM, "HELM",
+             rdkit_extensions::get_seq_extensions(Format::HELM)});
+        export_formats.push_back(
+            {Format::FASTA, "FASTA",
+             rdkit_extensions::get_seq_extensions(Format::FASTA)});
+    }
```
cdvonbargen: Is it wrong that it's weird to me that when you have monomers you can export HELM/FASTA _and_ the atomistic formats, but when you have atoms you can only export the atomistic formats?

I kinda want everything-all-ways (more complete) or only atom-formats-with-atomistic/monomer-formats-with-monomeric. What do people think?

KevKeating: > I kinda want everything-all-ways (more complete) or only atom-formats-with-atomistic/monomer-formats-with-monomeric. What do people think?

I could sort of see that.  We'd either require people to always explicitly convert between atomistic and monomeric, or the conversion is always allowed to be implicit.  There's a bit of a catch in that going from atomistic to monomeric doesn't necessarily have a correct answer (e.g. not everything with an `NCC=O` is a peptide), but presumably there'd be one obvious monomerization in the overwhelming majority of scenarios where people would actually _want_ to monomerize.

src/schrodinger/sketcher/dialog/file_import_export.cpp
cdvonbargen, line 53: Do we also need seq formats here so that we can import .helm and .fasta files?

src/schrodinger/sketcher/dialog/file_import_export.cpp
```diff
@@ -94,6 +94,15 @@ FormatList<Format> get_standard_export_formats()
         auto extensions = rdkit_extensions::get_mol_extensions(format);
         export_formats.push_back({format, label, extensions});
     }
+    // Sequence formats are always offered; rdkit_extensions::to_string handles
+    // atomistic <-> monomeric conversion on the fly, and conversion failures
+    // surface through the dialog's existing error path.
+    export_formats.push_back(
+        {Format::HELM, "HELM",
+         rdkit_extensions::get_seq_extensions(Format::HELM)});
```
cdvonbargen: I have a dumb question -- all 3 places where we use get_mol_extensions, we also add in get_seq_extensions -- should we just collapse those functions into get_mol_and_seq_extensions?

d-b-w (approved):
Looks good to me

PR #345, opened 2026-05-11
Sync rdkit extensions from mmshare

rachelnwalker:
- Linked Case: SHARED-12236, SHARED-12280, SHARED-12206, SHARED-12277, SHARED-11801, SHARED-12299


rachelnwalker:
I don't think we need @i-tub's recent commit that adds the executable

i-tub:
> I don't think we need @i-tub's recent commit that adds the executable

I agree; if we want to, we can have a simpler and better script to do the same thing in the core suite by using the Python API, which wasn't an option in the sketcher repo.

cdvonbargen (approved):

PR #347, opened 2026-05-11
SKETCH-2765: Prevent drifting of sugar-base bond lengths

JarrettSJohnson:
- Linked Case: SKETCH-2765

### Description
As nucleic acid chains grew, so did the sugar-base bond. Seems like existing monomers grew by roughly 0.12 units because all monomers in the nascent chain goes through the `resize_monomers` pass. Added a new function which initially sets the `MONOMER_ITEM_SIZE` atom property via `update_monomer_size` call. Now, subsequent calls to `updateMonomerLabelSizeOnModel` will check if the atom label has this already set and only perform displacement on those without it.

renamed `setMonomerSizes` to `resizeMonomers` so that it's clear that one function is responsible for the initial set, while another performs the update.

Additional change: also added const-refs to places where it made sense.

Adding @ZontaNicola since this seems like his wheelhouse.

### Testing Done
Created some tests to ensure prevention of bead-drifting as it grows.

### DIFF
--- a/src/schrodinger/rdkit_extensions/helm/monomer_coordgen.cpp
+++ b/src/schrodinger/rdkit_extensions/helm/monomer_coordgen.cpp
@@ -32,6 +32,7 @@
 
 #include <algorithm>
 #include <array>
+#include <cassert>
 #include <cmath>
 #include <iterator>
 #include <map>
@@ -3090,6 +3091,13 @@ void update_monomer_sizes(
     }
 }
 
+void store_initial_monomer_sizes(
+    RDKit::ROMol& mol,
+    const std::unordered_map<int, RDGeom::Point3D>& monomer_sizes)
+{
+    update_monomer_sizes(mol, monomer_sizes);
+}
+
 bool is_geometrically_regular_ring_2d(const RDKit::ROMol& mol,
                                       const std::vector<int>& ring_atoms,
                                       double radius_tol, double angle_tol)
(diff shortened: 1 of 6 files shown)


src/schrodinger/rdkit_extensions/helm/monomer_coordgen.cpp
```diff
@@ -3090,6 +3091,13 @@ void update_monomer_sizes(
     }
 }
 
+void store_initial_monomer_sizes(
+    RDKit::ROMol& mol,
+    const std::unordered_map<int, RDGeom::Point3D>& monomer_sizes)
+{
+    update_monomer_sizes(mol, monomer_sizes);
+}
```
ethan-schrodinger, line 3099: It looks like this function just wraps `update_monomer_sizes()`. We already extensively document the need for this call.

ethan-schrodinger (approved):

KevKeating (approved):

cdvonbargen:
@d-b-w  -- if you could sub in for Rachel while she's out -- once you approve Kevin can override the CODEOWNER merge req

src/schrodinger/rdkit_extensions/helm/monomer_coordgen.cpp
```diff
@@ -3186,13 +3194,25 @@ RingResizeInfo compute_ring_info_for_resize(RDKit::ROMol& mol)
...
+    RDKit::ROMol& mol,
+    const std::unordered_map<int, RDGeom::Point3D>& monomer_sizes)
 {
     if (monomer_sizes.empty()) {
         return;
     }
 
+    // callers must call store_initial_monomer_sizes first on any
+    // monomer that's never been sized. Otherwise the default-vs-actual size
+    // gap shows up here as a phantom resize and shifts neighbours.
```
d-b-w:
```suggestion
    // gap shows up here as a phantom resize and shifts neighbors.
```

America!

d-b-w (approved):
Great!

src/schrodinger/rdkit_extensions/helm/monomer_coordgen.cpp
JarrettSJohnson: 🇺🇸

ZontaNicola (approved):
Thanks for fixing this! I think we might want to get rid of the resizing altogether at some point and just regenerate coordinates only when we know the monomer sizes, but for now this is good.

PR #349, opened 2026-05-14
SKETCH-2745: Show corner wedge on selection when analogs are available

JarrettSJohnson:
- Linked Case: SKETCH-2745

### Description
Addressing UX feedback to show corner wedge on hover rather than just only on selection.

### Testing Done
Manually hovered over buttons with analogs

### DIFF
--- a/src/schrodinger/sketcher/widget/monomer_tool_widget.cpp
+++ b/src/schrodinger/sketcher/widget/monomer_tool_widget.cpp
@@ -186,27 +186,31 @@ MonomerToolWidget::MonomerToolWidget(QWidget* parent) :
 
     for (auto& entry : m_button_amino_acid_bimap.left) {
         auto* button = entry.first;
-        auto aa_tool = entry.second;
-        auto symbol = AMINO_ACID_TOOL_TO_RES_NAME.at(aa_tool);
-        auto it = analogs_by_aa.find(symbol);
-        if (it == analogs_by_aa.end() || it->second.empty()) {
-            continue;
-        }
-        const auto& name = AMINO_ACID_TOOL_TO_FULL_NAME.at(aa_tool);
-        auto* popup = new AminoAcidSymbolPopup(symbol, name, it->second, this);
         auto* modular_btn = qobject_cast<ModularToolButton*>(button);
         Q_ASSERT_X(modular_btn, "MonomerToolWidget",
                    "Expected ModularToolButton");
+        // Hide the wedge by default for every AA button; only enable hover
+        // reveal on buttons that actually get an analog popup attached.
+        modular_btn->showPopupIndicator(false);
+
+        auto aa_tool = entry.second;
+        auto symbol = AMINO_ACID_TOOL_TO_RES_NAME.at(aa_tool);
+        const auto& name = AMINO_ACID_TOOL_TO_FULL_NAME.at(aa_tool);
+        auto it = analogs_by_aa.find(symbol);
+        if (it == analogs_by_aa.end() || it->second.empty()) {
+            button->setToolTip(QString::fromStdString(name));
+            continue;
+        }
+        auto* popup = new AminoAcidSymbolPopup(symbol, name, it->second, this);
         modular_btn->setPopupWidget(popup);
         modular_btn->setEnumItem(0); // default to standard AA
-        modular_btn->showPopupIndicator(false);
+        modular_btn->showPopupIndicatorOnHover(true);
         m_amino_acid_symbol_popups[button] = popup;
     }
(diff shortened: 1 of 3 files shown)


src/schrodinger/sketcher/widget/monomer_tool_widget.cpp
```diff
@@ -186,27 +186,31 @@ MonomerToolWidget::MonomerToolWidget(QWidget* parent) :
 
     for (auto& entry : m_button_amino_acid_bimap.left) {
         auto* button = entry.first;
+        auto* modular_btn = qobject_cast<ModularToolButton*>(button);
+        Q_ASSERT_X(modular_btn, "MonomerToolWidget",
+                   "Expected ModularToolButton");
```
ethan-schrodinger, line 191: Was this meant for debugging, or is the intention to push this to production?

src/schrodinger/sketcher/widget/tool_button_with_popup.cpp
```diff
@@ -72,6 +75,16 @@ void ToolButtonWithPopup::showPopupIndicator(bool show)
     updateStyle();
 }
 
+void ToolButtonWithPopup::showPopupIndicatorOnHover(bool show)
+{
+    if (m_show_corner_arrow_on_hover == show) {
```
ethan-schrodinger: Nit: `showPopupIndicatorOnHover()` vs. `m_show_corner_arrow_on_hover`. Is there a reason to prefer "popup indicator" to "corner arrow" or vice versa? We should probably be consistent.

Additionally, should this function start with the word "set", since it appears to be a setter?

ethan-schrodinger (approved):

src/schrodinger/sketcher/widget/monomer_tool_widget.cpp
JarrettSJohnson: The original intention was to keep these in like we do for C `assert` as it's only fireable in debug mode. This was already in, and just moved from a previous line, but I don't mind removing these. There's still others left in the codebase, but if `Q_ASSERT`s are not desirable, I'll create a PR to remove them all that I've introduced already.

src/schrodinger/sketcher/widget/monomer_tool_widget.cpp
KevKeating, line 1: I was going to comment that you might be able to implement this with a style sheet along the lines of
```
QToolButton::menu-indicator { image: none; }
QToolButton::menu-indicator:hover { image: url(":icons/menu_corner_arrow.svg"); }
QToolButton::menu-indicator:checked { image: url(":icons/menu_corner_arrow.svg"); }
```
but I  just tested and that doesn't actually work.  It looks like there's no way to combine `::menu-indicator` and `:checked`, so the approach you used seems like the right one.

src/schrodinger/sketcher/widget/monomer_tool_widget.cpp
```diff
@@ -500,16 +513,22 @@ void MonomerToolWidget::onNucleicAcidClicked(QAbstractButton* button)
...
         auto* modular_btn = qobject_cast<ModularToolButton*>(button);
-        auto symbol = it->second->getSymbolForId(modular_btn->getEnumItem());
-        auto na_tool = m_button_nucleic_acid_bimap.left.at(button);
-        ping_or_set_model_value(getModel(), ModelKey::NUCLEIC_ACID_SYMBOL,
-                                NucleicAcidMutation{na_tool, symbol});
+        symbol = it->second->getSymbolForId(modular_btn->getEnumItem());
+    } else {
+        auto res_it = NUCLEIC_ACID_TOOL_TO_RES_NAME.find(na_tool);
+        if (res_it == NUCLEIC_ACID_TOOL_TO_RES_NAME.end()) {
+            return;
```
KevKeating, line 528: Is this just making sure that Sketcher doesn't crash if `NUCLEIC_ACID_TOOL_TO_RES_NAME` is missing a value?  A comment mentioning that would probably be helpful.

JarrettSJohnson: It's for guarding against full nucleotides that don't have a mapping. i've added a comment.

src/schrodinger/sketcher/widget/tool_button_with_popup.cpp
JarrettSJohnson: It's probably best to keep it similar to this because this function was made to be similar to the previously existing `void showPopupIndicator(bool show);`

KevKeating (approved):

PR #354, opened 2026-05-18
SHARED-12279: Kekulize when converting monomeric to SDF

i-tub:
- Linked Case: SHARED-12279

### Description

Updated rdkit_extensions.to_string() to force kekulization in the case of monomeric input and MDL output. We don't want aromatic bonds in SDF files because they can lead to kekulization errors later when read by RDKit, particularly when aromatic nitrogen is involved.

For atomistic inputs we still preserve the caller's representation.

### Testing Done

Just that it builds. There are unit tests in mmshare.


whosayn (approved):
@cdvonbargen this PR syncs mmshare changes to the sketcher OSS.

PR #356, opened 2026-05-19
SKETCH-2667: Using JavaScript's readText API to access the browser clipboard when Qt's clipboard access fails

KevKeating:
- Linked Case: SKETCH-2667

### Description

When Qt's clipboard access fails due to browser security, this PR allows the Sketcher to request access the clipboard via JavaScript's `readText` API, which should prompt the browser to request permission from the user.  I thought this was going to be straight-forward, but ran into a complication: When using asyncify, `await`ing a JavaScript promise in a Qt slot or event handler causes an immediate crash.  In other words `pasteAt` can't wait for the user to approve the clipboard access.  Instead, `pasteAt` has to return before waiting for the user, and there's now a JavaScript callback that will call C++ to complete the paste once Sketcher is given clipboard access.

With this change, pasting via the Sketcher menu bar should now work even if the user hasn't pasted via a keyboard shortcut first.  However, this won't work if the user tries to paste via the context menu in the Paste in Text... dialog.  The logic for that paste is buried pretty deep in a Qt private class (QWidgetTextControl), so there's no real ways to replace that code with the `readText` API.  The user can still paste into that dialog using Ctrl+V if they want, after which the dialog's context menu will work.

This change also works with Jarrett's changes for SKETCH-2751 (#341). I actually included his PR in my branch when I was working on this, but then backed it out at the end so that this PR wouldn't depend on that one.  I think there'll be a fairly trivial merge conflict in `SketcherWidget::getClipboardContents` (the `data == nullptr` check added here should, not surprisingly, go before the `data->hasFormat(SKETCHER_MIME_TYPE)` check) but the two changes work nicely together.

According to Claude (which wrote much of the code here), the `await`/Qt-slot-or-event-handler limitation doesn't exist in JSPI (which is basically asyncify version 2), but that's not even an option until we upgrade to Qt 6.10, which requires a fix or workaround for [QTBUG-145012](https://qt-project.atlassian.net/browse/QTBUG-145012).  Eventually, we may be able to remove some of this logic, though (or possibly all of it if Qt switches their WASM clipboard access to use the new API).

### Testing Done

Tested WASM builds using both Firefox and Chrome on Windows.  Since the build was hosted as localhost, which is allowed clipboard access by default, I didn't get any permission prompt, but pasting via the menu worked correctly even if I had pasted via a keyboard shortcut.  On Firefox, I get a weird additional button that needs to be clicked before the paste happens, but I don't think there's anything we can do about that.  (I think that's down to Firefox's implementation of `readText`.  I've seen that button on Firefox with other websites before.)

image

### DIFF
--- a/src/schrodinger/sketcher/sketcher_widget.cpp
+++ b/src/schrodinger/sketcher/sketcher_widget.cpp
@@ -534,6 +534,10 @@ void SketcherWidget::setInterfaceType(InterfaceTypeType interface_type)
 std::string SketcherWidget::getClipboardContents() const
 {
     auto data = QApplication::clipboard()->mimeData();
+    if (data == nullptr) {
+        // mimeData can return a nullptr in WASM builds
+        return "";
+    }
     if (data->hasText()) {
         return data->text().toStdString();
     }
@@ -545,6 +549,13 @@ void SketcherWidget::setClipboardContents(std::string text) const
     auto data = new QMimeData;
     data->setText(QString::fromStdString(text));
     QApplication::clipboard()->setMimeData(data);
+
+#ifdef __EMSCRIPTEN__
+    // Use the browser's aync clipboard api to enable copy for the wasm build
+    emscripten::val navigator = emscripten::val::global("navigator");
+    navigator["clipboard"].call<emscripten::val>("writeText",
+                                                 emscripten::val(text));
+#endif
 }
 
 void SketcherWidget::cut(Format format)
@@ -570,13 +581,6 @@ void SketcherWidget::copy(Format format, SceneSubset subset)
     }
     setClipboardContents(text);
     // SKETCH-2091: Add image content to the clipboard; blocked by SKETCH-1975
-
-#ifdef __EMSCRIPTEN__
-    // Use the browser's aync clipboard api to enable copy for the wasm build
-    emscripten::val navigator = emscripten::val::global("navigator");
-    navigator["clipboard"].call<emscripten::val>("writeText",
-                                                 emscripten::val(text));
-#endif
 }
 
 void SketcherWidget::copyAsImage()
@@ -589,6 +593,48 @@ void SketcherWidget::copyAsImage()
     QApplication::clipboard()->setImage(image);
 }
 
+#ifdef __EMSCRIPTEN__
+namespace
+{
+// State for the in-flight browser clipboard read kicked off by pasteAt(). The
+// .then() callback in sketcher_start_browser_clipboard_read re-enters via
+// sketcher_finish_browser_paste(), which uses these to complete the paste.
+SketcherWidget* g_pending_paste_widget = nullptr;
+std::optional<QPointF> g_pending_paste_position;
+} // namespace
+
+// Non-suspending: starts readText() and attaches a .then() that calls back
+// into C++. We must NOT await here -- ASYNCIFY cannot suspend through the JS
+// trampoline Qt uses for slot dispatch, so the read has to complete on a
+// fresh wasm stack (see sketcher_finish_browser_paste below).
+EM_JS(void, sketcher_start_browser_clipboard_read, (), {
+    navigator.clipboard.readText()
+        .then(function(text) {
+            const byteLength = lengthBytesUTF8(text) + 1;
+            const ptr = _malloc(byteLength);
+            stringToUTF8(text, ptr, byteLength);
+            _sketcher_finish_browser_paste(ptr);
+            _free(ptr);
+        })
+        .catch(function(err) {
+            // No text, permission denied, document not focused, etc.
+            _sketcher_finish_browser_paste(0);
+        });
+});
+
+extern "C" EMSCRIPTEN_KEEPALIVE void
+sketcher_finish_browser_paste(const char* text)
+{
+    auto* widget = g_pending_paste_widget;
+    auto position = g_pending_paste_position;
+    g_pending_paste_widget = nullptr;
+    g_pending_paste_position.reset();
+    if (widget && text && *text) {
+        widget->completePaste(text, position);
+    }
+}
+#endif
+
 /**
  * @internal
  * paste is agnostic of NEW_STRUCTURES_REPLACE_CONTENT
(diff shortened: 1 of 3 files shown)


ethan-schrodinger (approved):

cdvonbargen (approved):
This is.....kinda awful and makes me want to seriously consider a claude-d rewrite of the FE 😮‍💨

JarrettSJohnson:
(Late) Ack'ed by me. 

Just wanted to quickly mention that even when you go to Qt 6.10, I would not rely on JSPI just yet unless ya'll don't care about Firefox and Safari support (both of which are still experimental/enabled-only-in-nightly-builds in those browsers). (Currently also battling async issues in WebPyMOL)

KevKeating:
> (Late) Ack'ed by me.
> 
> Just wanted to quickly mention that even when you go to Qt 6.10, I would not rely on JSPI just yet unless ya'll don't care about Firefox and Safari support (both of which are still experimental/enabled-only-in-nightly-builds in those browsers). (Currently also battling async issues in WebPyMOL)

That's good to know!  I knew Safari was behind, but I didn't realize that Firefox was as well.  I think emscripten also generates a warning saying that JSPI is considered beta when you try to build anything with it enabled.

src/schrodinger/sketcher/sketcher_widget.cpp
```diff
@@ -589,16 +593,73 @@ void SketcherWidget::copyAsImage()
...
  * paste is agnostic of NEW_STRUCTURES_REPLACE_CONTENT
  */
 void SketcherWidget::pasteAt(std::optional<QPointF> position)
 {
     auto text = getClipboardContents();
-    if (text.empty()) {
+    if (!text.empty()) {
+        completePaste(std::move(text), position);
         return;
     }
```
juzerzarif, line 648: This might end up pasting "stale" clipboard contents into the sketcher. If you have:
1. Copy `CCCC` from the sketcher's canvas (sketcher's internal clipboard and system clipboard have `CCCC`)
2. Copy `CN` from somewhere else (system clipboard has `CN`, sketcher's internal clipboard still has `CCCC`)
3. Paste via context menu -> sketcher's internal clipboard returns `CCCC` so `text` isn't empty but it's also not the most recently copied thing

I think we're probably gonna have to always read from the system clipboard via JS in WASM land.

KevKeating: > This might end up pasting "stale" clipboard contents into the sketcher. If you have:
> 
> 1. Copy `CCCC` from the sketcher's canvas (sketcher's internal clipboard and system clipboard have `CCCC`)
> 
> 2. Copy `CN` from somewhere else (system clipboard has `CN`, sketcher's internal clipboard still has `CCCC`)
> 
> 3. Paste via context menu -> sketcher's internal clipboard returns `CCCC` so `text` isn't empty but it's also not the most recently copied thing

That's a good point!  I can reproduce the behavior you're describing.  I expected that step 2 would've cleared the Qt clipboard, but that's definitely not the case.  (I'd thought the Qt clipboard was always a view into the system clipboard, just sometimes an opaque and non-informative view.).  I can actually reproduce that behavior even without this PR.  Just replace step 3 with:

- Open the Paste in Text dialog
- Select paste from the context menu.  "CCCC" is pasted.
- Immediately hit Ctrl + V.  "CN" is pasted.
- Select paste from the context menu.  "CN" is pasted.

> I think we're probably gonna have to always read from the system clipboard via JS in WASM land.

That could lead to some behavior changes once https://github.com/schrodinger/sketcher/pull/341 is pushed, since that uses an alternate MIME_TYPE to store, e.g., coordinates for monomeric models, whereas the pure text version of the clipboard only contains a HELM string.  It looks like it is possible to put custom MIME types on the system clipboard from Javascript, so we might want to consider something like that.

juzerzarif: Hmmm, the JS clipboard API only supports putting certain supported types into the clipboard (https://developer.mozilla.org/en-US/docs/Web/API/ClipboardItem/supports_static - not every environment will support that "web " prefix thing). I wonder if we could do something like read the system clipboard and the Qt clipboard and if the text in both of them matches (same HELM string) we read the rdmol binary data from the Qt clipboard, otherwise we use the string from the system clipboard.
Though I don't know that much about HEML->RDMOL conversion so I don't know how resilient this would be. @JarrettSJohnson @rachelnwalker thoughts?

KevKeating: > I wonder if we could do something like read the system clipboard and the Qt clipboard and if the text in both of them matches (same HELM string) we read the rdmol binary data from the Qt clipboard, otherwise we use the string from the system clipboard.

Yeah, that had been my thought as well.  It could still potentially lead to issues*, although it seems vastly preferable to the current behavior.  For another option, what if we always did a `readText` call before checking the contents of the Qt clipboard?  That seems like it _should_ force the Qt clipboard to sync up with the system clipboard, which might solve our problem?  I can give that a try and see what happens.

&nbsp; 

\* If a user was to do something like:
- Open up Sketcher on two separate tabs
- Paste a HELM string into one of the tabs
- Modify the coordinates of the monomers
- Copy the modified coordinates
- Clear the Sketcher
- Paste the text into a local text editor
- Go do something else with their computer, possibly including modifying the clipboard contents
- Copy the HELM string from the text editor
- Paste into both Sketcher tabs

Then the paste would have different effects on the two tabs.  The tab where the copy happened would get the modified coordinates, and the other tab would get the default coordinates.  That seems like it would be fairly confusing to the user, but admittedly unlikely to actually occur in practice.

KevKeating: > For another option, what if we always did a readText call before checking the contents of the Qt clipboard? That seems like it should force the Qt clipboard to sync up with the system clipboard, which might solve our problem? I can give that a try and see what happens.

Nope, that doesn't work at all.  The Qt clipboard retains the old contents even after being given permission to access the system clipboard.

juzerzarif: Yeah from my poking around, it didn't seem like the Qt clipboard ever reads the system clipboard in WASM :(
> Then the paste would have different effects on the two tabs. The tab where the copy happened would get the modified coordinates, and the other tab would get the default coordinates. That seems like it would be fairly confusing to the user, but admittedly unlikely to actually occur in practice.

Yeah this is the bit I'm unsure about - I don't really know what a good UX would be for handling this and what's reasonably practical user behavior.

PR #357, opened 2026-05-20
SKETCH-2772: Anchor wheel zoom at the cursor

cdvonbargen:
- Linked Case: SKETCH-2772

### Description

Updates the zoom behavior (both pinch and wheel-zoom) to be focused at the cursor point.

### Testing Done

Added a test; manually tested with the trackpad on my mac

### DIFF
--- a/src/schrodinger/sketcher/molviewer/view.cpp
+++ b/src/schrodinger/sketcher/molviewer/view.cpp
@@ -168,11 +168,12 @@ void View::scaleAnchoredAtCursor(qreal scale_factor, const QPoint& anchor)
     // do it manually via sceneRect (the mechanism this view uses for
     // panning). Iterate so any sub-pixel residual from setSceneRect's
     // snapping converges to zero.
+    constexpr qreal SCENE_COORD_EPSILON = 0.001;
     auto scene_pos_before = mapToScene(anchor);
     scaleSafely(scale_factor);
     for (int i = 0; i < 4; ++i) {
         auto delta = scene_pos_before - mapToScene(anchor);
-        if (delta.manhattanLength() < 0.001) {
+        if (delta.manhattanLength() < SCENE_COORD_EPSILON) {
             break;
         }
         translateViewport(delta);
--- a/src/schrodinger/sketcher/molviewer/view.cpp
+++ b/src/schrodinger/sketcher/molviewer/view.cpp
@@ -113,9 +113,13 @@ void View::pinchTriggered(QPinchGesture* gesture)
     if (state == Qt::GestureStarted) {
         m_currently_pinching_trackpad = true;
     } else {
-        // zoom
+        // zoom anchored at the pinch center (the mouse cursor on a macOS
+        // trackpad pinch). Anchoring at the view center would push the
+        // structure away from where the user is pinching.
         auto scale_factor = gesture->scaleFactor();
-        scaleSafely(scale_factor);
+        auto pinch_anchor =
+            viewport()->mapFromGlobal(QCursor::pos());
+        scaleAnchoredAtCursor(scale_factor, pinch_anchor);
 
         auto angle = gesture->rotationAngle() - gesture->lastRotationAngle();
 
@@ -153,17 +157,22 @@ void View::wheelEvent(QWheelEvent* event)
     // K * 2^x, where K is an empirical constant to make the zooming feel
     // natural
     qreal scal = pow(2.0, event->angleDelta().y() / 2400.0);
-    // Capture the scene point under the cursor before the scale, then shift
-    // the viewport after so the same scene point is back under the cursor.
-    // Qt's setTransformationAnchor(AnchorUnderMouse) compensates via scroll
-    // bars, which this view has disabled, so do it via sceneRect (the
-    // mechanism this view already uses for panning) instead. Iterate so any
-    // sub-pixel residual from setSceneRect's snapping converges to zero.
-    auto cursor = event->position().toPoint();
-    auto cursor_scene_before = mapToScene(cursor);
-    scaleSafely(scal);
+    scaleAnchoredAtCursor(scal, event->position().toPoint());
+}
+
+void View::scaleAnchoredAtCursor(qreal scale_factor, const QPoint& anchor)
+{
+    // Capture the scene point under the anchor (cursor / pinch center) before
+    // scaling and shift the view afterwards so the same scene point is back
+    // under the anchor. Qt's setTransformationAnchor(AnchorUnderMouse)
+    // compensates through the scroll bars, which this view has disabled, so
+    // do it manually via sceneRect (the mechanism this view uses for
+    // panning). Iterate so any sub-pixel residual from setSceneRect's
+    // snapping converges to zero.
+    auto scene_pos_before = mapToScene(anchor);
+    scaleSafely(scale_factor);
     for (int i = 0; i < 4; ++i) {
-        auto delta = cursor_scene_before - mapToScene(cursor);
+        auto delta = scene_pos_before - mapToScene(anchor);
         if (delta.manhattanLength() < 0.001) {
             break;
         }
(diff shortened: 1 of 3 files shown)


src/schrodinger/sketcher/molviewer/view.h
```diff
@@ -130,6 +130,14 @@ class SKETCHER_API View : public QGraphicsView
      */
     void scaleSafely(qreal scale_factor);
 
+    /**
+     * Scale the matrix and shift the viewport so the scene point currently
+     * under @p anchor (in viewport coordinates) stays under @p anchor after
```
ethan-schrodinger, line 135: Is "@p" some well-known shorthand for the cursor location that I'm not familiar with?

test/schrodinger/sketcher/molviewer/test_view.cpp
```diff
@@ -51,5 +54,36 @@ BOOST_AUTO_TEST_CASE(test_delayed_fit_to_screen)
...
+ * SKETCH-2772: a wheel zoom must keep the scene point under the cursor fixed,
+ * so users can focus on a region by zooming while pointing at it.
+ */
+BOOST_AUTO_TEST_CASE(test_wheel_zoom_keeps_cursor_point_fixed)
+{
+    auto scene = TestScene::getScene();
+    TestView view(scene.get());
+    view.resize(400, 400);
+    view.show();
+    [[maybe_unused]] auto exposed = QTest::qWaitForWindowExposed(&view);
```
ethan-schrodinger, line 67: I guess this is a nit, but since we don't care about the return value of this function, why bother assigning it to a variable?

ethan-schrodinger (approved):

src/schrodinger/sketcher/molviewer/view.h
JarrettSJohnson: `@p` short for `@param`--Claude loves this style.

src/schrodinger/sketcher/molviewer/view.cpp
```diff
@@ -153,7 +156,28 @@ void View::wheelEvent(QWheelEvent* event)
...
+    // compensates through the scroll bars, which this view has disabled, so
+    // do it manually via sceneRect (the mechanism this view uses for
+    // panning). Iterate so any sub-pixel residual from setSceneRect's
+    // snapping converges to zero.
+    constexpr qreal SCENE_COORD_EPSILON = 0.001;
+    auto scene_pos_before = mapToScene(anchor);
+    scaleSafely(scale_factor);
+    for (int i = 0; i < 4; ++i) {
+        auto delta = scene_pos_before - mapToScene(anchor);
+        if (delta.manhattanLength() < SCENE_COORD_EPSILON) {
```
JarrettSJohnson, line 176: Curious about the rationale of using `manhattanLength` here and not euclidean?

src/schrodinger/sketcher/molviewer/view.cpp
```diff
@@ -153,7 +156,28 @@ void View::wheelEvent(QWheelEvent* event)
...
+    // scaling and shift the view afterwards so the same scene point is back
+    // under the anchor. Qt's setTransformationAnchor(AnchorUnderMouse)
+    // compensates through the scroll bars, which this view has disabled, so
+    // do it manually via sceneRect (the mechanism this view uses for
+    // panning). Iterate so any sub-pixel residual from setSceneRect's
+    // snapping converges to zero.
+    constexpr qreal SCENE_COORD_EPSILON = 0.001;
+    auto scene_pos_before = mapToScene(anchor);
+    scaleSafely(scale_factor);
+    for (int i = 0; i < 4; ++i) {
```
JarrettSJohnson, line 174: I assume `4` is just a heuristic for convergence? as opposed to just using `while (true)`?

JarrettSJohnson (approved):
👍

src/schrodinger/sketcher/molviewer/view.cpp
```diff
@@ -113,9 +113,12 @@ void View::pinchTriggered(QPinchGesture* gesture)
     if (state == Qt::GestureStarted) {
         m_currently_pinching_trackpad = true;
     } else {
-        // zoom
+        // zoom anchored at the pinch center (the mouse cursor on a macOS
```
KevKeating, line 116: Should this be "a trackpad" instead of "a macOS trackpad"?  Is something about this specific to MacOS?

src/schrodinger/sketcher/molviewer/view.h
KevKeating: I found this confusing as well.

src/schrodinger/sketcher/molviewer/view.cpp
KevKeating, line 174: Is this iteration necessary because `anchor` is a `QPoint` instead of a `QPointF` (and because `mapToScene` only has a `QPoint` implementation, not a `QPointF` one, so `anchor` would be implicitly rounded even if it was a QPointF)?  In other words, would we avoid this if we/Qt handled floating point values properly here?

(I'm not saying it's necessarily worth changing even if the answer is "yes."  I'm just trying to make sure I understand why we need this.)

src/schrodinger/sketcher/molviewer/view.cpp
```diff
@@ -113,9 +113,12 @@ void View::pinchTriggered(QPinchGesture* gesture)
...
+        // trackpad pinch). Anchoring at the view center would push the
+        // structure away from where the user is pinching.
         auto scale_factor = gesture->scaleFactor();
-        scaleSafely(scale_factor);
+        auto pinch_anchor = viewport()->mapFromGlobal(QCursor::pos());
```
KevKeating, line 120: It would probably be better to use `gesture.centerPoint()` here so we're pulling all of our data from the event itself instead of the state of the system when this method is run (and it would give us a QPointF instead of a QPoint).

test/schrodinger/sketcher/molviewer/test_view.cpp
```diff
@@ -51,5 +54,36 @@ BOOST_AUTO_TEST_CASE(test_delayed_fit_to_screen)
     BOOST_TEST(view.m_initial_geometry_set);
 }
 
+/**
+ * SKETCH-2772: a wheel zoom must keep the scene point under the cursor fixed,
+ * so users can focus on a region by zooming while pointing at it.
+ */
+BOOST_AUTO_TEST_CASE(test_wheel_zoom_keeps_cursor_point_fixed)
+{
+    auto scene = TestScene::getScene();
```
KevKeating, line 63: We should probably set `Qt::WA_DontShowOnScreen` in the `TestView` constructor so it doesn't actually get displayed.

test/schrodinger/sketcher/molviewer/test_view.cpp
KevKeating: +1

src/schrodinger/sketcher/molviewer/view.cpp
ZontaNicola: general question: what happens on a touchscreen pinch? do we want to use the midpoint of the pinch action if we're pinching on the screen and the cursor position if we're pinching on a trackpad?

src/schrodinger/sketcher/molviewer/view.cpp
cdvonbargen: Claude defended itself since either norm works near zero. Where euclidean isn't a one-shot call like this.

src/schrodinger/sketcher/molviewer/view.cpp
cdvonbargen: yeah, claude is wary of a something pathalogical; it extracted out MAX_CORRECTIONS

cdvonbargen: Claude defended itself -- "Partly — mapToScene only has a QPoint overload, so there's inherent integer truncation regardless of the anchor type" -- so I had it add a comment.

src/schrodinger/sketcher/molviewer/view.cpp
cdvonbargen: Switched so that it also addresses the touchscreen

PR #359, opened 2026-05-22
SKETCH-2378: Use getSGroupDataLabels() for DAT SGroup label positioning

cdvonbargen:
- Linked Case: SKETCH-2378

### Description

Replaces the manual FIELDDISP parsing in `SGroupItem` with the new `MolDraw2D_detail::getSGroupDataLabels()` API added in [rdkit/rdkit#9189](https://github.com/rdkit/rdkit/pull/9189). The old code ignored the absolute/relative positioning flag and always treated FIELDDISP coordinates as relative to the first atom; the new API properly handles both cases along with edge cases like missing atoms or FIELDDISP entries.

**Build blocker:** The pre-built RDKit package does not install `MolDraw2DDetails.h` as a public header, and the `RDKit::MolDraw2D` cmake target has broken freetype paths from the build bot. This PR needs either: (a) the header added to the RDKit install set and the cmake target fixed, or (b) a local RDKit source build.

### Testing Done

Not yet compilable due to the RDKit packaging issue described above. Once resolved, verify with:
```
buildinger sketcher -i --with-tests --test-args="test/schrodinger/sketcher/molviewer/test_sgroup_item"
```


cdvonbargen:
@d-b-w -- when I did https://github.com/rdkit/rdkit/pull/9189/changes I didnt realize that MolDraw2DDetails.h would n't be accessible -- what is the right thing to do here?

PR #360, opened 2026-05-23
SKETCH-2375: Migrate to RDKit native attachment point handling

cdvonbargen:
- Linked Case: SKETCH-2375

### RDKit APIs used

- `MolOps::expandAttachmentPoints()` replaces `molattachpt_property_to_attachment_point_dummies()`.
- `MolOps::collapseAttachmentPoints()` replaces `attachment_point_dummies_to_molattachpt_property()`.
- `MolOps::details::addExplicitAttachmentPoint()` replaces `make_new_attachment_point()`.

### Functions kept for Sketcher behavior

- `is_attachment_point_dummy()` cannot use RDKit's `isAttachmentPoint()` because RDKit tests whether an AP can be collapsed and therefore rejects wedged APs. Sketcher must still recognize those as attachment points and support legacy `_AP<n>` labels.
- `compute2DCoords()` additionally handles HELM structures, frozen atoms, ring templates, and removal of stale molfile wedging.
- `wedgeMolBonds()` prevents RDKit from choosing attachment-point bonds for wedging by temporarily making those bonds ineligible.
- The Sketcher R-group helpers for AP numbering, renumbering, bond/atom lookup, and counting implement Sketcher's global `_AP<n>` UI numbering. RDKit's `_fromAttachPoint` values only describe the first or second attachment on an individual parent atom.
- MDL export clears the bond direction on attachment-point bonds before collapsing, since RDKit won't collapse a wedged AP and the uncollapsed dummy would otherwise be written as a bare `*` that reads back as an ordinary dummy atom.

### Testing Done

Added new tests; all existing tests pass

### DIFF
--- a/src/schrodinger/rdkit_extensions/rgroup.cpp
+++ b/src/schrodinger/rdkit_extensions/rgroup.cpp
@@ -11,7 +11,6 @@
 #include <stdexcept>
 
 #include <rdkit/GraphMol/Atom.h>
-#include <rdkit/GraphMol/MolOps.h>
 #include <rdkit/GraphMol/QueryAtom.h>
 
 #include "schrodinger/rdkit_extensions/constants.h"
@@ -21,11 +20,6 @@ namespace schrodinger
 {
 namespace rdkit_extensions
 {
-
-bool is_attachment_point_dummy(const RDKit::Atom& atom)
-{
-    return RDKit::MolOps::details::isAttachmentPoint(&atom);
-}
 
 [[nodiscard]] std::shared_ptr<RDKit::Atom>
 make_new_r_group(const unsigned int r_group_num)
--- a/src/schrodinger/rdkit_extensions/rgroup.cpp
+++ b/src/schrodinger/rdkit_extensions/rgroup.cpp
@@ -23,9 +23,16 @@ namespace rdkit_extensions
 
 bool is_attachment_point_dummy(const RDKit::Atom& atom)
 {
+    if (atom.getAtomicNum() != 0 || atom.getTotalDegree() != 1) {
+        return false;
+    }
+    int from_ap{0};
+    if (atom.getPropIfPresent(RDKit::common_properties::_fromAttachPoint,
+                              from_ap)) {
+        return from_ap == 1 || from_ap == 2;
+    }
     std::string label;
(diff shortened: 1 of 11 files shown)


src/schrodinger/rdkit_extensions/rgroup.cpp
```diff
@@ -21,12 +21,18 @@ namespace schrodinger
...
+    if (atom.hasProp(RDKit::common_properties::_fromAttachPoint)) {
+        return true;
+    }
+
     std::string label;
-    return atom.getAtomicNum() == 0 && atom.getTotalDegree() == 1 &&
-           atom.getPropIfPresent(RDKit::common_properties::atomLabel, label) &&
-           label.find(ATTACHMENT_POINT_LABEL_PREFIX) == 0;
+    return atom.getPropIfPresent(RDKit::common_properties::atomLabel, label) &&
+           label.starts_with(ATTACHMENT_POINT_LABEL_PREFIX);
```
ethan-schrodinger, line 35: Under what conditions would we see the attachment point prefix on the atom label, but we wouldn't see the `_fromAttachPoint` property?

ethan-schrodinger (commented):
No "testing done" section. I assume this was tested?

src/schrodinger/rdkit_extensions/rgroup.cpp
cdvonbargen: _fromAttachPoint is only set by RDKit when reading SD files or using the  (relatively new) addExplicitAttachmentPoint/expandAttachmentPoints functions directly; so things like CXSMILES wouldn't have it.

ethan-schrodinger: Understood, thanks. Is this well known, or would we benefit from adding a comment here to this effect?

ethan-schrodinger (approved):

rachelnwalker (approved):
>is_attachment_point_dummy() cannot use RDKit's isAttachmentPoint() because RDKit tests whether an AP can be collapsed and therefore rejects wedged APs

doesn't that mean that exporting to SDF will have different behavior dependent on whether you have a wedged bond? Normally you'll get ATTCHPT=1 on the parent, but if the bond is wedged you'd just get a dummy atom? I'm not sure what sort of issues this would cause - if our internal tests pass though, I'm good with merging this

PR #365, opened 2026-06-01
SKETCH-2667: Copy and paste in WASM builds no longer uses QClipboard and preserves binary RDMol

KevKeating:
- Linked Case: SKETCH-2667
 
### Description

I think I finally have a good solution for copying and pasting in WASM builds.  With this PR, desktop app copying is unchanged, but we no longer use QClipboard at all in WASM.  Instead, we only use the system clipboard, so we shouldn't have to worry about anything getting out of sync.  To accomplish this, web copying now uses either 2 or 3 different MIME types, depending on the browser:

| MIME type | Browser | Example data |
|-----------|---------|--------------|
| text/plain | all |  PEPTIDE1{A.A}$$$$V2.0 |
| text/html  | all | &lt;div data-x-schrodinger-sketcher="KLUv/WDcAc0KAPZSQjswxTiNATDkH0rVAE8KY4FSAwMTNhAIMt01v3sCMmw1G06KeKidinkL9NCbrfhPSKwT6F1UN/kbjITIFCwANQAwAAfW/vvA+QI0YAoPNiJ5ZFj2Xzz/WoT+lbjzI6t/L5y35N/f+vo1/wFQVAQEktuoBGVFebGrn6xUpV0DoYqCGYisIqpJN037M1v+kwQL1ACg/md8IkxWKlYpFgTDvwLCZQ/S62b4Y+SMHL8l5MlULE1a418wwe/HZ7NXCRt1mDRTCdv4Dfb6d0UZSaT9GPnR0QwLmz2nzsTpFzisX0ez+snJbbFqHDA2IV88TVpDXgQ7bltU2F2gnPSjJm77IqgRdn+GpiYJKgweAF9LhAi5BuUyCkhrAGxShe8CZJc3SOA8APUGsmnMUa7GoLz1tWQp5I0mGwl5XQaYejdsEgDMDdWwF928gTe4CeyrFU19JcacWRyAEg=="> &lt;/div>PEPTIDE1{A.A}$$$$V2.0 |
| web&nbsp;application&#8725;x&#8209;schrodinger&#8209;sketcher | recent Chromium-based | KLUv/WDcAc0KAPZSQjswxTiNATDkH0rVAE8KY4FSAwMTNhAIMt01v3sCMmw1G06KeKidinkL9NCbrfhPSKwT6F1UN/kbjITIFCwANQAwAAfW/vvA+QI0YAoPNiJ5ZFj2Xzz/WoT+lbjzI6t/L5y35N/f+vo1/wFQVAQEktuoBGVFebGrn6xUpV0DoYqCGYisIqpJN037M1v+kwQL1ACg/md8IkxWKlYpFgTDvwLCZQ/S62b4Y+SMHL8l5MlULE1a418wwe/HZ7NXCRt1mDRTCdv4Dfb6d0UZSaT9GPnR0QwLmz2nzsTpFzisX0ez+snJbbFqHDA2IV88TVpDXgQ7bltU2F2gnPSjJm77IqgRdn+GpiYJKgweAF9LhAi5BuUyCkhrAGxShe8CZJc3SOA8APUGsmnMUa7GoLz1tWQp5I0mGwl5XQaYejdsEgDMDdWwF928gTe4CeyrFU19JcacWRyAEg==  |

Basically, `text/plain` contains the user-requested format in plain text, as it did before.  The `web application/x-schrodinger-sketcher` data contain the base64-encoded binary rdmol data, but this requires that the browser supporting custom web MIME types, which is currently only available on newer Chromium-based browsers.  (It's an accepted web standard and Firefox is [planning](https://bugzilla.mozilla.org/show_bug.cgi?id=1956304) on implementing it eventually, but it's relatively newish.)  The `text/html` data contains both the plain text and the binary rdmol data.  The rdmol is stored as a data attribute of an empty `div` tag, which means that anything other than Sketcher will ignore that data and only see the plain text, since some applications use `text/html` data to paste rich text.

When pasting, we first look for `web application/x-schrodinger-sketcher` data.  If that's not present, we look at `text/html` data and see if it contains any binary data.  If not, we use the `text/plain` data.  That way, copying from, e.g., Firefox to Firefox or between Firefox and Chrome will still transfer the binary data, and copying from Chrome to Chrome avoids the overhead of going through HTML.  I'm not worried about the CPU usage of the copy, but there are a bunch of behind-the-scenes operations that seem best to avoid wherever possible.  For example, browsers typically add additional HTML tags to the copied data, such as `<html><p>...</p></html>` or `<meta>` tags containing meta-data.  That's why `sketcher_start_browser_clipboard_read` needs to use a `DOMParser` instead of simple string matching to pull data out of the `<div>` block; the data that gets returned from the clipboard is different from the data that we put there in the first place.  Additionally, browsers do some sanitization and security checks to `text/html` data to make sure that it doesn't contain any malicious scripts.  I wouldn't expect anything we do to trigger any issues, but it's a bunch of additional moving parts that we avoid by using `web application/x-schrodinger-sketcher` in browsers that support it.

### Testing Done

I tested copies and pastes in and between Firefox and Chrome on Windows, and I used https://evercoder.github.io/clipboard-inspector/ to manually check clipboard contents.  I also sanity checked copy and paste on the desktop app.

### DIFF
--- a/src/schrodinger/sketcher/sketcher_widget.cpp
+++ b/src/schrodinger/sketcher/sketcher_widget.cpp
@@ -577,8 +577,8 @@ EM_JS(void, sketcher_write_clipboard,
             .replace(/&/g, '&amp;')
             .replace(/</g, '&lt;')
             .replace(/>/g, '&gt;');
-        const html = '<div data-' + appName + '="' + binary + '"></div>' +
-                     escapeHtml(text);
+        const html = '<div style="display: none;" data-' + appName + '="' +
+                    binary + '"></div>' + escapeHtml(text);
         items['text/html'] = new Blob([html], {type: 'text/html'});
         if (sketcher_browser_supports_web_mime(web_mime_ptr)) {
             items[webMime] = new Blob([binary], {type: webMime});
--- a/src/schrodinger/sketcher/sketcher_widget.cpp
+++ b/src/schrodinger/sketcher/sketcher_widget.cpp
@@ -19,7 +19,6 @@
 
 #ifdef __EMSCRIPTEN__
 #include <emscripten.h>
-#include <emscripten/val.h>
 #endif
 
 #include "schrodinger/rdkit_extensions/helm.h"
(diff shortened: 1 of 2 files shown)


src/schrodinger/sketcher/sketcher_widget.cpp
```diff
@@ -533,40 +532,98 @@ void SketcherWidget::setInterfaceType(InterfaceTypeType interface_type)
...
+    const appName = UTF8ToString(app_name_ptr);
+    const items = {
+        'text/plain': new Blob([text], {type: 'text/plain'}),
+    };
+    if (binary.length > 0) {
+        const escapeHtml = (s) => s
+            .replace(/&/g, '&amp;')
+            .replace(/</g, '&lt;')
+            .replace(/>/g, '&gt;');
+        const html = '<div data-' + appName + '="' + binary + '"></div>' +
```
juzerzarif: Maybe we should also add a `style="display: none;"` attribute to prevent the div from affecting layout as well

juzerzarif (approved):
This is way nicer than what I had in mind!

src/schrodinger/sketcher/sketcher_widget.cpp
KevKeating: Good idea!  Just added it.

ethan-schrodinger (approved):
Learned a lot from this. :)

PR #367, opened 2026-06-05
SKETCH-2763: Generate complement strand for nucleic acids

JarrettSJohnson:
- Linked Case: SKETCH-2763

### Description

This revision offers support for the generation of a complement strand for a selected set of nucleic acids.

Adds a menu option to generate the complement whenever a set of nucleic acids are selected; mixed selections disables the option.

Ordering and pointer invalidation is a bit tricky for building up the sequences in this way, so we employ an index-based strategy to build up the nascent chain for each polymer.

### Testing Done

Adds some tests for conditions where menus should show, generation of complement, and undo.

### DIFF
--- a/src/schrodinger/sketcher/menu/monomer_context_menu.h
+++ b/src/schrodinger/sketcher/menu/monomer_context_menu.h
@@ -47,6 +47,11 @@ class SKETCHER_API MonomerContextMenu : public AbstractContextMenu
     void mutateMonomerRequested(std::vector<MonomerMutation> mutations,
                                 QString description);
 
+    // `bases` holds raw atom pointers into the live molecule, so the receiver
+    // must consume them before it starts mutating the molecule.
+    void addComplementaryStrandRequested(
+        const std::unordered_set<const RDKit::Atom*>& bases);
+
   protected:
     void updateActions() override;
 
@@ -56,6 +61,7 @@ class SKETCHER_API MonomerContextMenu : public AbstractContextMenu
     void createProtonateAction();
     void createMutateBaseSubMenu();
     void createSugarToggleAction();
+    void createAddComplementaryStrandAction();
     void createDeleteAction();
 
     QMenu* m_mutate_residue_menu = nullptr;
@@ -63,6 +69,12 @@ class SKETCHER_API MonomerContextMenu : public AbstractContextMenu
     QAction* m_protonate_action = nullptr;
     QMenu* m_mutate_base_menu = nullptr;
     QAction* m_sugar_toggle_action = nullptr;
+    QAction* m_add_complement_action = nullptr;
+
+    // Bases from the current selection that have a Watson-Crick complement.
+    // Computed once per updateActions() and reused both to set the action's
+    // enabled state and as the payload when the action is triggered.
+    std::unordered_set<const RDKit::Atom*> m_complement_bases;
 };
 
 } // namespace sketcher
--- a/src/schrodinger/sketcher/model/mol_model.cpp
+++ b/src/schrodinger/sketcher/model/mol_model.cpp
@@ -2,6 +2,8 @@
 
 #include <algorithm>
 #include <functional>
+#include <limits>
+#include <map>
 #include <variant>
 
 #include <fmt/format.h>
@@ -9,6 +11,7 @@
 #include <rdkit/GraphMol/Atom.h>
 #include <rdkit/GraphMol/Bond.h>
 #include <rdkit/GraphMol/Conformer.h>
+#include <rdkit/GraphMol/MonomerInfo.h>
 #include <rdkit/GraphMol/MolOps.h>
 #include <rdkit/GraphMol/RWMol.h>
 #include <rdkit/GraphMol/SubstanceGroup.h>
@@ -28,6 +31,7 @@
 
 #include "schrodinger/rdkit_extensions/helm.h"
 #include "schrodinger/rdkit_extensions/molops.h"
+#include "schrodinger/rdkit_extensions/monomer_database.h"
 #include "schrodinger/rdkit_extensions/monomer_mol.h"
 #include "schrodinger/rdkit_extensions/rgroup.h"
 #include "schrodinger/sketcher/rdkit/sgroup.h"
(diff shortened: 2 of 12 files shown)


src/schrodinger/sketcher/menu/monomer_context_menu.h
```diff
@@ -47,6 +47,12 @@ class SKETCHER_API MonomerContextMenu : public AbstractContextMenu
     void mutateMonomerRequested(std::vector<MonomerMutation> mutations,
                                 QString description);
 
+    // Must be wired with a direct (synchronous) connection — `bases`
+    // holds raw atom pointers that are invalidated as soon as the
+    // receiver starts mutating the molecule.
```
KevKeating: I get why the pointers are invalidated, but why does that mean that the connection must be direct?  I would think that we can safely assume that nothing is going to mutate the molecule in between this signal being emitted and the slot being called even with a queued connection.

src/schrodinger/sketcher/model/mol_model.cpp
```diff
@@ -171,6 +175,108 @@ void strip_notes_and_mol_model_tags(RDKit::ROMol& mol)
     }
 }
 
+// Per-base data needed to build one complement monomer triplet (sugar, base,
+// phosphate). `orig_idx` is the index of the original base it pairs with.
```
KevKeating: This and the static method docstrings should use the
```
/**
 *
 */
```
comment formatting.

src/schrodinger/sketcher/model/mol_model.cpp
```diff
@@ -171,6 +175,108 @@ void strip_notes_and_mol_model_tags(RDKit::ROMol& mol)
...
+        if (base == nullptr || get_monomer_type(base) != MonomerType::NA_BASE) {
+            continue;
+        }
+        bases_by_polymer[rdkit_extensions::get_polymer_id(base)].push_back(
+            base->getIdx());
+    }
+    return bases_by_polymer;
+}
+
+// Resolve each base in `base_idxs` to its complement triplet, in 5'->3' order.
```
KevKeating: I think it would be clearer to describe the ordering here as "residue number order" or something similar instead of "5' -> 3'" order.  Based on this docstring, I was wondering if the function examined connectivity to determine order.

src/schrodinger/sketcher/model/mol_model.cpp
```diff
@@ -171,6 +175,108 @@ void strip_notes_and_mol_model_tags(RDKit::ROMol& mol)
...
+        if (!comp.has_value()) {
+            continue;
+        }
+        // Phosphate is always "P"; sugar identity (R vs dR) is per base to
+        // handle chimeric chains mixing R and dR within one polymer.
+        std::string sugar_sym = is_dna ? "dR" : "R";
+        if (!db.getMonomerSmiles(sugar_sym, rdkit_extensions::ChainType::RNA) ||
+            !db.getMonomerSmiles(*comp, rdkit_extensions::ChainType::RNA)) {
+            continue;
+        }
```
KevKeating: Is there any reason to check the monomer DB here?  We know that `sugar_sym` and the residue names returned by `get_complement_base_symbol` are valid monomer names, so this check doesn't seem to add much.  Same with the check on line 212.

src/schrodinger/sketcher/model/mol_model.cpp
```diff
@@ -171,6 +175,108 @@ void strip_notes_and_mol_model_tags(RDKit::ROMol& mol)
...
+    return pairs;
+}
+
+// Direction from an original base toward its complement: away from the base's
+// own sugar, so the complement follows a rotated/moved strand. Derived from the
+// first base with a locatable sugar; falls back to straight down. Also fills
+// `orig_base_positions` with each pair's base position, in order.
+static RDGeom::Point3D
+pairing_axis(const RDKit::ROMol& mol, const std::vector<ComplementPair>& pairs,
+             std::vector<RDGeom::Point3D>& orig_base_positions)
```
KevKeating: I think it might be clearer to have a separate function for populating `orig_base_positions`.  It doesn't seem like there's any substantial savings in having both operations as part of the same function, and the call to this function in `addComplementChainForPolymer` looks like it just forgot to initialize `orig_base_positions`.  The change should also make this function a fair bit simpler.

src/schrodinger/sketcher/model/mol_model.cpp
```diff
@@ -939,6 +1045,119 @@ void MolModel::addMonomericConnection(const RDKit::Atom* const monomer_one,
...
+    // invalidates any held conformer reference.
+    std::vector<RDGeom::Point3D> orig_base_positions;
+    const RDGeom::Point3D pair_dir =
+        pairing_axis(m_mol, pairs, orig_base_positions);
+    // Backbone axis: perpendicular to the pairing axis (-x for the canonical
+    // downward pair_dir).
+    const RDGeom::Point3D backbone_dir(pair_dir.y, -pair_dir.x, 0.0);
+    // Lay the complement out one monomer bond at a time so the spacing matches
+    // the rest of the monomer layout.
+    constexpr double MONOMER_STEP = rdkit_extensions::MONOMER_BOND_LENGTH;
```
KevKeating: Is this constant just to reduce the line length relative to using `rdkit_extensions::MONOMER_BOND_LENGTH`?  It seems like the `MONOMER_BOND_LENGTH` name is more descriptive/readable.

src/schrodinger/sketcher/model/mol_model.cpp
```diff
@@ -939,6 +1045,119 @@ void MolModel::addMonomericConnection(const RDKit::Atom* const monomer_one,
...
+        // further out so the new base faces back toward the original.
+        // Phosphate offsets along the backbone axis.
+        RDGeom::Point3D base_coord =
+            along(orig_base_pos, pair_dir, MONOMER_STEP);
+        RDGeom::Point3D sugar_coord = along(base_coord, pair_dir, MONOMER_STEP);
+        RDGeom::Point3D phos_coord =
+            along(sugar_coord, backbone_dir, MONOMER_STEP);
+
+        if (is_first) {
+            addMonomer(pair.sugar_sym, rdkit_extensions::ChainType::RNA,
```
KevKeating: It looks like this function is going to add one item to the undo stack per monomer.  You could use `createUndoMacro` to avoid this.

src/schrodinger/sketcher/model/mol_model.cpp
```diff
@@ -939,6 +1045,119 @@ void MolModel::addMonomericConnection(const RDKit::Atom* const monomer_one,
...
+    // after round-tripping through the parser.
+    auto renumber = [this, new_chain_atom_idxs]() {
+        for (size_t i = 0; i < new_chain_atom_idxs.size(); ++i) {
+            auto* atom = m_mol.getAtomWithIdx(new_chain_atom_idxs[i]);
+            auto* info = dynamic_cast<RDKit::AtomPDBResidueInfo*>(
+                atom->getMonomerInfo());
+            if (info == nullptr) {
+                continue;
+            }
+            info->setResidueNumber(static_cast<int>(i + 1));
```
KevKeating, line 1195: Is this using one residue number per monomer?  Each residue should consist of a sugar, base, and phosphate.

src/schrodinger/sketcher/model/mol_model.cpp
JarrettSJohnson: There is already one above in the caller of this function via:

```cpp
    // One undo entry covers the complement chains for every source polymer.
    auto undo_macro = createUndoMacro("Add Complementary Sequence");
    for (const auto& [polymer_id, base_idxs] : bases_by_polymer) {
        addComplementChainForPolymer(base_idxs);
    }
}
```
Is this not fine?

src/schrodinger/sketcher/model/mol_model.cpp
JarrettSJohnson: I modeled this after 

https://github.com/schrodinger/sketcher/blob/e48a7c787609d7a73206946b115974cd9252eaa2/src/schrodinger/rdkit_extensions/helm/to_rdkit.cpp#L272-L278

which seems to assign per monomer rather than actual residue--unless I'm missing something obvious?

src/schrodinger/sketcher/model/mol_model.cpp
KevKeating: Ahh, I missed that.  Never mind then!

src/schrodinger/sketcher/model/mol_model.cpp
KevKeating: That's a good question.  I wrote [get_residue_number_for_new_monomer](https://github.com/schrodinger/sketcher/blob/7a2748a0406eedb0e062f1dd95e55c2e6e5b5a85/src/schrodinger/sketcher/model/mol_model.cpp#L755) on the assumption that residue numbering for HELM nucleic acids should follow standard residue numbering rules, but that's not what happens in `to_rdkit`:
```py
>>> rdmol = rdkit_extensions.to_rdkit("RNA1{R(U)P.R(T)P}$$$$V2.0")
>>> [atom.GetMonomerInfo().GetResidueNumber() for atom in rdmol.GetAtoms()]
[1, 2, 3, 4, 5, 6]
```
@d-b-w and/or @rachelnwalker (or anyone else who knows), is that intentional?  How should HELM residue numbering work for nucleic acids?

rachelnwalker: Yeah this is intentional because the HELM reader/writer assumes each 'monomer' is one 'residue' 

Definitely strange, but I don't think it would be straightforward to change that - I think we'd need to add a different property to differentiate nucleotides if you need that

KevKeating: Ahh, okay.  I don't need anything for differentiating nucleotides currently.  We might eventually need a property for "what's the residue number of the Maestro structure residue that this monomer is mapped to," but adding a property for that wouldn't be hard.  I'll file a case to modify get_residue_number_for_new_monomer so that it follows this behavior.

KevKeating: I filed SKETCH-2782 to modify `get_residue_number_for_new_monomer`.  @JarrettSJohnson, you can ignore my original comment here!

rachelnwalker: >We might eventually need a property for "what's the residue number of the Maestro structure residue that this monomer is mapped to,"

Yup that makes sense

src/schrodinger/sketcher/model/mol_model.cpp
```diff
@@ -171,6 +175,137 @@ void strip_notes_and_mol_model_tags(RDKit::ROMol& mol)
     }
 }
 
+/**
+ * @brief Per-base data needed to build one complement monomer triplet (sugar,
+ * base, phosphate).
+ *
+ * @c orig_idx is the index of the original base this triplet pairs with.
+ */
+struct ComplementPair {
```
ethan-schrodinger: Nit: maybe I'm missing something, but I don't know whether `ComplementPair` is the best name for this, since it's not a pair of anything.

src/schrodinger/sketcher/model/mol_model.cpp
```diff
@@ -171,6 +175,137 @@ void strip_notes_and_mol_model_tags(RDKit::ROMol& mol)
...
+    size_t orig_idx;
+    std::string sugar_sym;
+    std::string comp_sym;
```
ethan-schrodinger: `comp_sym` = "complement base symbol" ?

Could we choose a more descriptive variable name? Or at least add a @param field to make this explicit?

src/schrodinger/sketcher/rdkit/monomeric.cpp
```diff
@@ -130,6 +130,51 @@ std::string get_monomer_res_name(const RDKit::Atom* const monomer)
...
+    // The purine A pairs with the partner strand's pyrimidine: T in DNA, U in
+    // RNA. Both pyrimidines T and U pair back to A.
+    if (base_symbol == "A") {
+        return is_dna ? "T" : "U";
+    }
+    if (base_symbol == "T" || base_symbol == "U") {
+        return "A";
+    }
+    return std::nullopt;
+}
```
ethan-schrodinger, line 176: Definitely a nit, but I guess it would probably be marginally easier to implement two separate functions (or even just maps) for DNA vs. RNA. The caller already has to supply `is_dna`, so it would just then also be responsible for choosing the correct function to call. Then, the implementation would be as simple as accessing a map element in each case. Alternatively, of course, we could just access the map element directly from the calling code.

ethan-schrodinger (approved):
LGTM!

I have some minort comments, but obviously it's up to you whether they are worth implementing.

src/schrodinger/sketcher/model/mol_model.cpp
JarrettSJohnson: Fixed — replaced `comp_sym` with the descriptive `base_symbol` field and use `complement_base_symbol` for the lookup result.

Signed Codex

src/schrodinger/sketcher/rdkit/monomeric.cpp
JarrettSJohnson: Fixed — split the lookup into `get_dna_complement_base_symbol` and `get_rna_complement_base_symbol`, each backed by its own map. The caller now selects the appropriate function based on `is_dna`, and focused unit coverage was added.

Signed Codex

src/schrodinger/sketcher/model/mol_model.cpp
JarrettSJohnson: Fixed — renamed `ComplementPair` to `ComplementNucleotide` and renamed its fields to `original_base_idx`, `sugar_symbol`, and `base_symbol` so the type describes the single complement nucleotide it represents.

Signed Codex

src/schrodinger/sketcher/model/mol_model.cpp
```diff
@@ -940,6 +1083,122 @@ void MolModel::addMonomericConnection(const RDKit::Atom* const monomer_one,
...
+    const std::unordered_set<const RDKit::Atom*>& selected_bases)
+{
+    auto bases_by_polymer = group_bases_by_polymer(selected_bases);
+    if (bases_by_polymer.empty()) {
+        return;
+    }
+    // One undo entry covers the complement chains for every source polymer.
+    auto undo_macro = createUndoMacro("Add Complementary Sequence");
+    for (const auto& [polymer_id, base_idxs] : bases_by_polymer) {
+        addComplementChainForPolymer(base_idxs);
```
KevKeating, line 1096: Will this add an empty entry to the undo stack if none of the polymers have any complementable nucleotides?

src/schrodinger/sketcher/model/mol_model.cpp
```diff
@@ -940,6 +1083,122 @@ void MolModel::addMonomericConnection(const RDKit::Atom* const monomer_one,
...
+            orig_base_pos, pair_dir, rdkit_extensions::MONOMER_BOND_LENGTH);
+        RDGeom::Point3D sugar_coord =
+            along(base_coord, pair_dir, rdkit_extensions::MONOMER_BOND_LENGTH);
+        RDGeom::Point3D phos_coord = along(
+            sugar_coord, backbone_dir, rdkit_extensions::MONOMER_BOND_LENGTH);
+
+        if (is_first) {
+            addMonomer(complement.sugar_symbol,
+                       rdkit_extensions::ChainType::RNA, sugar_coord);
+        } else {
```
KevKeating, line 1154: What happens if there's a non-complementable nucleotide in between two complementable nucleotides?  Will this create an extra long connection between the two complements?

src/schrodinger/sketcher/model/mol_model.cpp
```diff
@@ -171,6 +175,145 @@ void strip_notes_and_mol_model_tags(RDKit::ROMol& mol)
...
+ *
+ * Atom indices are stored (not pointers) since pointers are invalidated by
+ * later edits. Null/non-base atoms are skipped. An ordered map keeps per-
+ * polymer processing deterministic, so the auto-numbered complement chains are
+ * reproducible.
+ *
+ * @param bases the selected atoms
+ * @return a map from polymer id to the contained base atom indices
+ */
+static std::map<std::string, std::vector<size_t>>
```
KevKeating, line 202: These static methods and the ComplementNucleotide struct could probably go in monomeric.h/cpp since they don't seem to need access to anything MolModel-related.  That would keep this file a bit cleaner, but I'm okay either way.  We can always move them later if we need to access them somewhere else.

KevKeating (approved):
I had a few minor questions, but they can always be addressed later if need be.

PR #368, opened 2026-06-05
SKETCH-2754: Bound monomeric attachments points are now labeled on hover with the select, erase, and move tools

KevKeating:
- Linked Case: SKETCH-2754

### Description

This PR moves the logic for labeling the attachment points of monomeric connectors from AbstractMonomerSceneTool to StandardSceneToolBase.  As a result, these labels will appear when using the select, erase, and move tools.  This required a bit of work to detangle the logic for labeling bound attachment points on monomeric connectors from the logic for labeling unbound attachment points on monomers (which is still in AbstractMonomerSceneTool).  Note that the diff here is _quite_ large, but there's no actual new logic; just a lot of moved and refactored code.

Because the labeling logic is now in StandardScenToolBase, these bound attachment point labels will also appear with the atomistic tools once those tools can be active with a monomer present in the workspace.  That seems like it's the right behavior.  The monomeric connections get labels but no predictive highlight (unless the tool is already set to highlight monomeric connections), so there's no implication that the tool would be able interact with the connections.  If anyone thinks this might be confusing to the user, let me know.  It wouldn't be hard to put the attachment point labeling stuff in it's own AbstractSceneToolWithMonomerLabels subclass, which would let us control where it's active.

There are also a few quasi-related changes here that were required for this refactoring.  (These are why this diff touches so many files.)

- All scene tool constructors now take a Fonts object (since we need fonts to generate the labels)
- AbstractHighlightingItem::highlightItems now has two overloads, one that takes a list of `QGraphicsItem*`s and one that takes a list of `const QGraphicsItem*`s.  (`QGraphicsItem*` can be implicitly coerced into a `const QGraphicsItem*`, but that conversion doesn't apply to _containers_ of pointers, so we need an explicit overload.)  I've almost needed this a handful of times and here it was required to avoid a const cast, so I had Claude do the tedious work of implementing this.

### Testing Done

All existing unit tests pass, and I confirmed that the appropriate labels appear with the select, move, and erase tools.

### DIFF
--- a/src/schrodinger/sketcher/tool/abstract_monomer_scene_tool.cpp
+++ b/src/schrodinger/sketcher/tool/abstract_monomer_scene_tool.cpp
@@ -51,6 +51,7 @@ void AbstractMonomerSceneTool::updateColorsAfterBackgroundColorChange(
 
 void AbstractMonomerSceneTool::onStructureUpdated()
 {
+    StandardSceneToolBase::onStructureUpdated();
     // the unbound attachment point items themselves were children of their
     // respective monomers, so Qt destroyed them when the monomers were
     // destroyed. We just need to clear out the stale pointers here.
--- a/src/schrodinger/sketcher/tool/abstract_monomer_scene_tool.cpp
+++ b/src/schrodinger/sketcher/tool/abstract_monomer_scene_tool.cpp
@@ -1,19 +1,13 @@
 #include "schrodinger/sketcher/tool/abstract_monomer_scene_tool.h"
 
-#include <algorithm>
-#include <cmath>
 #include <vector>
 
 #include <QGraphicsItem>
-#include <QPainterPath>
 
 #include "schrodinger/sketcher/molviewer/abstract_monomer_item.h"
-#include "schrodinger/sketcher/molviewer/monomer_attachment_point_labels.h"
 #include "schrodinger/sketcher/molviewer/monomer_connector_item.h"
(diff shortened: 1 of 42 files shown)


src/schrodinger/sketcher/tool/abstract_monomer_scene_tool.cpp
```diff
@@ -66,31 +41,22 @@ AbstractMonomerSceneTool::~AbstractMonomerSceneTool()
...
     bool is_dark_mode)
 {
     StandardSceneToolBase::updateColorsAfterBackgroundColorChange(is_dark_mode);
     m_unbound_ap_label_color =
         is_dark_mode ? UNBOUND_AP_LABEL_COLOR_DARK_BG : UNBOUND_AP_LABEL_COLOR;
-    m_bound_ap_label_color =
-        is_dark_mode ? BOUND_AP_LABEL_COLOR_DARK_BG : BOUND_AP_LABEL_COLOR;
 }
 
 void AbstractMonomerSceneTool::onStructureUpdated()
```
JarrettSJohnson, line 52: Does this function need to also call its base class' method `DrawSceneToolBase::onStructureUpdated`? Not sure if the new `m_hovered_monomer_item` would be properly null'd out if not.

KevKeating: Yeah, it definitely should.  I'll fix that.

ethan-schrodinger (approved):

PR #370, opened 2026-06-09
SKETCH-2775: Warn instead of fail if Arimo font is present but can't be loaded

KevKeating:
- Linked Case: SKETCH-2775

### Description

Claude figured out what was going on here: Windows has a security setting that can cause font loads like this to fail, and it's frequently enabled on Enterprise installations (although it's available on Windows Home and Pro as well).  This PR modifies `load_font_resources` to warn instead of error when this happens.  The function will still fail if the font is completely missing, which is the main thing that the check was intended to catch, since a missing font means that there's something wrong with the build.

We could potentially do something more to try to get Arimo loaded in this scenario, like locally install it to the user's fonts directory, which would avoid triggering the security check when the font is loaded.  My first thought, though, is that it's probably not worth worrying about something like that for what appears to be a fairly uncommon scenario.

Here's Claude's full explanation:

> The failure is almost certainly a Windows kernel font-blocking security policy that Enterprise-managed machines commonly enable and that Home/Pro do not.
> 
> What's happening
> QFontDatabase::addApplicationFont on the Windows platform plugin ultimately calls the Win32 GDI function AddFontMemResourceEx (it loads the resource into memory, then registers it). It returns -1 when that underlying call fails.
> 
> The relevant policy is "Untrusted Font Blocking" (a.k.a. the PROCESS_MITIGATION_FONT_DISABLE_POLICY / "Block untrusted fonts" mitigation), introduced in Windows 10 to mitigate kernel GDI font-parsing exploits. When enabled, win32k.sys refuses to load any font that isn't installed in the protected system fonts directory (%windir%\Fonts). A font loaded from application memory — exactly what addApplicationFont does with your embedded .ttf — counts as "untrusted" and is blocked, so the call fails and Qt returns -1.
> 
> This is why it's version-correlated but not really about the version: Enterprise SKUs are the ones that get hardened via Group Policy / Intune / security baselines, where this mitigation is frequently turned on. The same code runs fine on an unmanaged Win11 Home/Pro box because the policy is off by default.
> 
> How to confirm
> Check the policy: Computer Configuration → Administrative Templates → System → Mitigation Options → Untrusted Font Blocking, or the registry key HKLM\SYSTEM\CurrentControlSet\Control\Session Manager\Kernel\MitigationOptions (the font-disable bits), or HKLM\...\Kernel\MitigationAuditOptions for audit mode.
> Enterprise machines may also surface this in the System event log (font-blocking events) when in audit/enforce mode.
> It can also be applied per-process via Exploit Protection ("Disable extension points" / font settings) in Windows Security.
> Possible mitigations (for your code)
> Don't hard-fail. This is the most defensible fix regardless of root cause. Throwing std::runtime_error takes the whole app down because a font couldn't register. Arimo is a Liberation/metric-compatible substitute for Arial — degrading to the system font is far better than crashing. Log a warning and continue.
> Force a non-GDI font backend. Qt's Windows plugin can use DirectWrite/FreeType instead of GDI (e.g. QT_QPA_PLATFORM=windows:fontengine=freetype, depending on Qt version). DirectWrite-based application font loading isn't subject to the GDI win32k mitigation, so this often sidesteps the block entirely. Worth testing on an affected machine.
> Install the fonts properly (per-user font install, no admin needed since Win10 1809) so they live in a trusted location — but that's heavier and changes deployment.
> I'd recommend (1) unconditionally and (2) if you actually need Arimo to render. Want me to rework load_font_resources() so a blocked font logs a warning and falls back instead of throwing?

### Testing Done

I enabled PROCESS_MITIGATION_FONT_DISABLE_POLICY for Sketcher.exe on my machine, which caused Sketcher launch to fail without this fix: I got a stack trace when I tried to launch the Sketcher.  With the fix, Sketcher launched correctly with warnings printed to the terminal:
```
Failed to register font (resource data present): :/resources/fonts/Arimo-Regular.ttf
Failed to register font (resource data present): :/resources/fonts/Arimo-Bold.ttf
Failed to register font (resource data present): :/resources/fonts/Arimo-Italic.ttf
Failed to register font (resource data present): :/resources/fonts/Arimo-BoldItalic.ttf
```


src/schrodinger/sketcher/font_loader.cpp
```diff
@@ -21,9 +23,22 @@ void load_font_resources()
...
+                fmt::format("Font resource not found: {}", font_path));
+        }
+        if (QFontDatabase::addApplicationFont(font_path) == -1) {
+            // The font data is present but the OS refused to register it. This
+            // happens on machines (commonly Windows enterprise installations)
+            // where a security policy blocks loading fonts that aren't
+            // installed in the protected system fonts directory. In this
+            // scenario, warn and fall back to a system font.
+            qWarning() << "Failed to register font (resource data present):"
+                       << font_path;
```
ethan-schrodinger, line 41: This implementation looks okay to me, but the nature of this bug raises a new question: just how sensitive is Sketcher appearance to the font being used? I know that in several places we require information about the font in order to calculate exact positioning, sizing, etc. Should we be concerned that using the system font instead of Arimo may cause problems? Is this worth certifying for if we expect this to be a common setting for Windows users (who comprise the majority of our installations)?

ethan-schrodinger (approved):

src/schrodinger/sketcher/font_loader.cpp
KevKeating: I'm hoping that the fact that we've only had one complaint so far means that it's not a very common setting, although I suppose it's also likely that not all of our customers have upgraded to 26-2 yet, so it's hard to gauge.

If Arimo couldn't be loaded, then any calculations would be done using the font chosen to replace it (since that's what would wind up initializing the QFontMetricsF object), so I wouldn't expect any serious issues due to the calculation differences.  I suppose my main concern would be that Qt would pick something serifed to replace Arimo (like what was happening in SKETCH-2728 / SKETCH-2682) since Qt wouldn't know enough about Arimo to pick a suitable replacement for it.  One reasonably easy option to deal with that would be to replace our [FONT_NAME](https://github.com/schrodinger/sketcher/blob/c8fdd5e98f18f307a51f8a7be3ccacfc6be58f5b/src/schrodinger/sketcher/molviewer/constants.h#L71) constant with a QStringList like [SVG_FONT_LIST](https://github.com/schrodinger/sketcher/blob/c8fdd5e98f18f307a51f8a7be3ccacfc6be58f5b/src/schrodinger/sketcher/molviewer/constants.h#L75).  That way, rendering in Sketcher would fall back to Arial, which should be available on any Windows machine, if loading Arimo fails.  There are some differences between the two fonts, but they're pretty minimal.  That change seems worth the small amount of effort required to implement it.  I can update this PR with that.

ethan-schrodinger (approved):

PR #372, opened 2026-06-11
SKETCH-2782: get_residue_number_for_new_monomer now uses one residue number per monomer, even for nucleic acids

KevKeating:
- Linked Case: SKETCH-2782

### Description

Based on [this discussion](https://github.com/schrodinger/sketcher/pull/367#discussion_r3364753980), I've modified `get_residue_number_for_new_monomer` to use one residue number per monomer, even for nucleic acids.

### Testing Done

Added unit tests coverage for the new logic.

### DIFF
--- a/src/schrodinger/sketcher/model/mol_model.cpp
+++ b/src/schrodinger/sketcher/model/mol_model.cpp
@@ -745,11 +745,20 @@ int get_residue_number_for_new_monomer(
     auto existing_res_nums =
         get_all_residue_numbers_in_polymer(bound_to_monomer);
 
-    // first, determine what the ideal residue number would be based on the
-    // residue number of the monomer we're binding to
+    // First, determine what the "ideal" residue number would be based on the
+    // offset from the residue number of the monomer we're binding to. For
+    // example, the ideal residue number of a new C-terminal peptide is one
+    // higher than the bound peptide, and the ideal residue number of a new
+    // N-terminal peptide is one lower than the bound peptide. We refer to it as
+    // an "ideal" residue number since we don't know whether we'll actually be
+    // able to use that number; it may be non-positive, which isn't allowed, or
+    // it may already be taken by an existing monomer. In those cases, we'll
+    // just use one more than the highest current residue number, since we know
+    // that's available and valid.
+
     int res_num_offset = 1;
     if (chain_type == ChainType::PEPTIDE &&
-        new_monomer_ap_name == ap_model_name_for(PeptideAP::N)) {
+        new_monomer_ap_name == ap_model_name_for(PeptideAP::C)) {
         // a new C terminal peptide residue
         res_num_offset = -1;
     } else if (chain_type == ChainType::RNA) {
@@ -765,22 +774,23 @@ int get_residue_number_for_new_monomer(
                 // a new sugar bound to its base
                 res_num_offset = -1;
             }
-        } else if (monomer_type == MonomerType::NA_BASE &&
-                   new_monomer_ap_name == ap_model_name_for(NA_BASE_AP_N1_9)) {
-            res_num_offset = -1;
-        } else if (monomer_type == MonomerType::NA_PHOSPHATE &&
-                   new_monomer_ap_name ==
-                       ap_model_name_for(NAPhosphateAP::TO_PREV_SUGAR)) {
-            res_num_offset = -1;
+        } else if (monomer_type == MonomerType::NA_PHOSPHATE) {
+            if (new_monomer_ap_name ==
+                ap_model_name_for(NAPhosphateAP::TO_PREV_SUGAR)) {
+                res_num_offset = 2;
+            } else if (new_monomer_ap_name ==
+                       ap_model_name_for(NAPhosphateAP::TO_NEXT_SUGAR)) {
+                res_num_offset = -1;
+            }
         }
     }
     auto bound_to_res_num =
         rdkit_extensions::get_residue_number(bound_to_monomer);
     int new_res_num = bound_to_res_num + res_num_offset;
 
-    // if the ideal residue number is not available, just use one more than the
-    // highest current residue number, since we know that's available
-    if (new_res_num < 0 || existing_res_nums.contains(new_res_num)) {
+    if (new_res_num <= 0 || existing_res_nums.contains(new_res_num)) {
+        // the ideal residue number isn't valid or isn't available. Note that
+        // residue numbers of 0 confuse HELM generation, so we avoid them.
         new_res_num = *std::ranges::max_element(existing_res_nums) + 1;
     }
     return new_res_num;
--- a/src/schrodinger/sketcher/model/mol_model.cpp
+++ b/src/schrodinger/sketcher/model/mol_model.cpp
@@ -733,11 +733,11 @@ get_all_residue_numbers_in_polymer(const RDKit::Atom* const atom_in_polymer)
     return residue_numbers;
 }
 
-int
-get_residue_number_for_new_monomer(const std::string_view res_name,
-                                   const rdkit_extensions::ChainType chain_type,
-                                   const std::string_view new_monomer_ap_name,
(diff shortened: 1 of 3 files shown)


src/schrodinger/sketcher/model/mol_model.cpp
```diff
@@ -728,67 +727,59 @@ get_all_residue_numbers_of_monomer_type_in_polymer(
...
-            new_monomer_ap_name == ap_model_name_for(NASugarAP::THREE_PRIME)) {
-            // sugars can link to the previous residue
+    } else if (chain_type == ChainType::RNA) {
+        if (monomer_type == MonomerType::NA_SUGAR) {
+            if (new_monomer_ap_name ==
+                ap_model_name_for(NASugarAP::THREE_PRIME)) {
+                // a new sugar bound to the next phosphate. We skip a number
+                // since the base is typically given the number immediately
+                // after the sugar
+                res_num_offset = -2;
```
ethan-schrodinger, line 771: Should we be worried about this "typically"?

src/schrodinger/sketcher/model/mol_model.cpp
```diff
@@ -728,67 +727,59 @@ get_all_residue_numbers_of_monomer_type_in_polymer(
...
 
     auto monomer_type = get_monomer_type_from_chain_type(chain_type, res_name);
-    auto existing_res_nums = get_all_residue_numbers_of_monomer_type_in_polymer(
-        monomer_type, bound_to_monomer);
+    auto existing_res_nums =
+        get_all_residue_numbers_in_polymer(bound_to_monomer);
 
+    // first, determine what the ideal residue number would be based on the
+    // residue number of the monomer we're binding to
     int res_num_offset = 1;
```
ethan-schrodinger: It might be good here to briefly explain what this offset value means, and when we should expect it to be positive and negative (the way we already do in the unit test).

src/schrodinger/sketcher/model/mol_model.cpp
```diff
@@ -728,67 +727,59 @@ get_all_residue_numbers_of_monomer_type_in_polymer(
...
-            res_num_offset = 0;
         }
     }
     auto bound_to_res_num =
         rdkit_extensions::get_residue_number(bound_to_monomer);
     int new_res_num = bound_to_res_num + res_num_offset;
+
+    // if the ideal residue number is not available, just use one more than the
+    // highest current residue number, since we know that's available
     if (new_res_num < 0 || existing_res_nums.contains(new_res_num)) {
```
ethan-schrodinger: We reference this concept is several places, but I don't know if there is a comment that explicitly states what makes this number "ideal"?

ethan-schrodinger (approved):

src/schrodinger/sketcher/model/mol_model.cpp
```diff
@@ -728,67 +727,59 @@ get_all_residue_numbers_of_monomer_type_in_polymer(
...
-                                   const RDKit::Atom* const bound_to_monomer)
+int get_residue_number_for_new_monomer(
+    const std::string_view res_name,
+    const rdkit_extensions::ChainType chain_type,
+    const std::string_view new_monomer_ap_name,
+    const RDKit::Atom* const bound_to_monomer)
 {
     using rdkit_extensions::ChainType;
 
     auto monomer_type = get_monomer_type_from_chain_type(chain_type, res_name);
```
rachelnwalker, line 744: General comment but not urgent - I'm not sure whether you will always be able to determine whether something is a base/sugar/phosphate based off residue name. I'm not sure how common it will be but there are customers with modified phosphate linkers and sugars that won't adhere to a certain naming convention - you maybe need to deduce by a substructure match, or just rely completely on the attachment points used

src/schrodinger/sketcher/model/mol_model.cpp
```diff
@@ -728,67 +727,59 @@ get_all_residue_numbers_of_monomer_type_in_polymer(
...
+                res_num_offset = -1;
+            }
+        } else if (monomer_type == MonomerType::NA_BASE &&
+                   new_monomer_ap_name == ap_model_name_for(NA_BASE_AP_N1_9)) {
+            res_num_offset = -1;
+        } else if (monomer_type == MonomerType::NA_PHOSPHATE &&
+                   new_monomer_ap_name ==
+                       ap_model_name_for(NAPhosphateAP::TO_PREV_SUGAR)) {
             res_num_offset = -1;
-        } else if (!((monomer_type == MonomerType::NA_PHOSPHATE &&
```
rachelnwalker, line 779: Wouldn't these offsets be +1 and +2? Totally possible I am not understanding the MonomerType enum correctly - is this a base connecting to a sugar then a phosphate connecting to the same sugar? if the sugar is 0, I would think that the base is 1 then the phosphate is 2

src/schrodinger/sketcher/model/mol_model.cpp
KevKeating: Not really.  This would only be an issue if the user is intentionally leaving an abasic site (i.e. a missing base), and even then, the only result of this would be that the residues aren't numbered sequentially, which shouldn't have any real consequences.  If we want a guarantee that we have the ideal residue numbering, i.e. starting at one and continuing sequentially in the correct order, we'll need to renumber the entire strand after it's been built.  That's probably not worth worrying about unless we know of something downstream that's requires it.

src/schrodinger/sketcher/model/mol_model.cpp
KevKeating: Yeah, agreed.  We might want to record information about the monomer type when the monomer is added, since Sketcher and/or the HELM converter would presumably be in the best position to figure that out (and the Sketcher would likely have additional information about the monomer based on the tool being used).  I've filed SKETCH-2796 for this.

src/schrodinger/sketcher/model/mol_model.cpp
KevKeating: Sure.  I've added a long comment explaining this and what "ideal" means.

src/schrodinger/sketcher/model/mol_model.cpp
KevKeating: Added.  See my response above.

src/schrodinger/sketcher/model/mol_model.cpp
KevKeating: Yeah, it looks like a bunch of the numbers here are off.  Good catch!  I'll fix that and update the tests, both to check for the correct results and hopefully to make those types of issues more obvious.  (I also temporarily tweaked monomer drawing locally so it includes the residue numbers to verify that the current numbering is wrong, since I kept getting myself massively confused trying to reason through the numbering here.)

test/schrodinger/sketcher/model/test_mol_model.cpp
```diff
@@ -4801,124 +4802,121 @@ BOOST_AUTO_TEST_CASE(test_addBoundMonomer_RNA)
...
+
+    // check residue numbers for sugars bound to the phosphate
+    auto prev_sugar_num = get_residue_number_for_new_monomer(
+        "R", ChainType::RNA, ap_model_name_for(NASugarAP::THREE_PRIME),
+        phosphate);
+    auto next_sugar_num = get_residue_number_for_new_monomer(
+        "R", ChainType::RNA, ap_model_name_for(NASugarAP::FIVE_PRIME),
+        phosphate);
+    BOOST_TEST(prev_sugar_num == 8u);
+    BOOST_TEST(next_sugar_num == 11u);
```
ethan-schrodinger, line 4900: I think it would be worth adding comments here (and on the test below) that briefly explain why we should expect these specific residue numbers.

ethan-schrodinger (approved):

test/schrodinger/sketcher/model/test_mol_model.cpp
KevKeating: Yeah, that seems helpful.  Added.

PR #373, opened 2026-06-15
Bump macOS to version 26

JarrettSJohnson:
- Linked Case: NOJIRAID

### Description

Was trying to build sketcher on macos (M4). Hit this error:
```
In file included from /Users/runner/work/sketcher/sketcher/build/external/qt-prefix/src/qt-build/qtbase/include/QtCore/qyieldcpu.h:1:
/Users/runner/work/sketcher/sketcher/build/external/qt-prefix/src/qt/qtbase/src/corelib/thread/qyieldcpu.h:35:5: error: implicitly declaring library function '__yield' with type 'void ()' [-Werror,-Wimplicit-function-declaration]
   35 |     __yield();              // Generic
      |     ^
/Users/runner/work/sketcher/sketcher/build/external/qt-prefix/src/qt/qtbase/src/corelib/thread/qyieldcpu.h:35:5: note: include the header <arm_acle.h> or explicitly provide a declaration for '__yield'
1 error generated.
ninja: build stopped: subcommand failed.
```
also reproducible when I bumped to `macos-26` on CI--did this since the XCode version reflects what we use for 26.3. Let me know if you dont want to include the bump to 26 here and just have the patch only.

Claude's explanation:

====

Apple clang 21+ has `__has_builtin(__yield)` as `true`, so Qt 6.x's qyieldcpu.h takes its generic `__yield()` branch — but `__yield()` is only declared in <arm_acle.h>, which that header never includes. Qt compiles with -Werror=implicit-function-declaration, so the implicit declaration becomes a hard error.

=====

Not sure if this may be addressed/solved with a more recent Qt?

### Testing Done
Tests on macos-26 pass.


KevKeating (approved):
> Let me know if you dont want to include the bump to 26 here and just have the patch only.

Bumping the builder seems reasonable to me.

> Not sure if this may be addressed/solved with a more recent Qt?

Anything newer than Qt 6.7 has deal-breaker WASM bugs, so this patch is probably the right solution.

cdvonbargen (approved):
I'm fine with this, but worried that it leaves us uncovered for macos-15 builds which I think is still how the core suite is built... Let's talk about it at standup next week

PR #376, opened 2026-06-18
SKETCH-2643: Migrate squish tests to CI via playwright tests

JarrettSJohnson:
- Linked Case: SKETCH-2643
- Primary Reviewer: @KevKeating

### Summary

Migrates the portable parts of the Squish 2D sketcher coverage into Playwright/WASM e2e tests.

- Adds fixture-backed import smoke coverage for historically problematic SDF/MAE structures from `suite_2D_sketcher/tst_problematic_structures`.
- Ports the reusable import-format coverage from `suite_2D_sketcher_new/tst_import_menu`, including methylphenidate fixtures across supported text formats, InChI/SMARTS/complex-SMILES import cases, and the oversized-structure rejection path.
- Ports the SHARED-6974 miscellaneous MAE regression structures.
- Adds behavioral coverage for keyboard element shortcuts that can be validated through exported SMILES.
- Extends the Playwright helpers with `clearSketcher` and `tryImport` so import sweeps can collect failures without aborting on the first thrown exception.

### Notes

Maestro-specific Squish behavior was intentionally left out, including Project Table workflows, file dialogs, Maestro help/export menus, and Qt pixel-reference checks that do not map cleanly to the standalone WASM sketcher.


Sean-Seekins:
This looks promising!  I just have one suggestion to make sure we go in the right direction.  We should use the most recent squish tests for this conversion: https://opengrok.schrodinger.com/xref/gui_tests/priority_4/suite_molviewer/.  There are "suite_2D_sketcher_new/" tests, which are for the 'new sketcher', as opposed to the 'old sketcher', which has tests in "https://opengrok.schrodinger.com/xref/gui_tests/priority_1/suite_2D_sketcher/".  

So the timeline went 'old sketcher' (priority_1/suite_2D_sketcher/) -> 'new sketcher' (priority_4/suite_2D_sketcher_new/) -> 'molviewer' (priority_4/suite_molviewer/), and now we have the monomeric sketcher as well.  The molviewer tests are stored in priority_4 so they don't run on automated runs (too tedious to manage those failures with image differences on different machines, so I just run locally).  

If that seems confusing, I agree.  It is because of the way that it is.

JarrettSJohnson:
> This looks promising! I just have one suggestion to make sure we go in the right direction. We should use the most recent squish tests for this conversion: https://opengrok.schrodinger.com/xref/gui_tests/priority_4/suite_molviewer/. There are "suite_2D_sketcher_new/" tests, which are for the 'new sketcher', as opposed to the 'old sketcher', which has tests in "https://opengrok.schrodinger.com/xref/gui_tests/priority_1/suite_2D_sketcher/".
> 
> So the timeline went 'old sketcher' (priority_1/suite_2D_sketcher/) -> 'new sketcher' (priority_4/suite_2D_sketcher_new/) -> 'molviewer' (priority_4/suite_molviewer/), and now we have the monomeric sketcher as well. The molviewer tests are stored in priority_4 so they don't run on automated runs (too tedious to manage those failures with image differences on different machines, so I just run locally).
> 
> If that seems confusing, I agree. It is because of the way that it is.

AFAICT, these are inspired from all of the above including `priority_4/suite_molviewer`. Let me know if there's anything I'm missing from there. They're not all yet ported due to some inability to query atom/bond coordinates, but that's seems to be flag'd on my agent's next TODO.

PR #379, opened 2026-06-24
SHARED-12481: Handle monomer upload errors more gracefully

ricrogz:
- Linked Case: SHARED-12481

### Description
This changes the behavior of `insert_monomers_from_json()`: instead of throwing when there a problem while adding one monomer definition to the DB, the error now gets logged. Note that errors associated with creating the DB, the DB schema, or parsing the json input will still throw an exception.

As a consequence, `MonomerDatabase::loadMonomersFromJson()` and `MonomerDatabase::insertMonomersFromJson()` will now return a `std::vector<std::string>`. The vector contains one log string for each of the monomers parsed from the input JSON. The log string can be either "Monomer inserted successfully {monomer details}" for a monomer that was correctly added to the DB, or, in case of a failure, a string with the description of the error followed by the data of the monomer being processed (see the test added to `test/schrodinger/rdkit_extensions/test_monomer_database.cpp` for some error string examples).


KevKeating (commented):
Is it worth logging the successes?  If we skipped those log entries, it would let the calling code determine whether the operation was completely successful by just checking .empty() on the returned vector.  That would also make it easier for the calling code to present only the failure messages to the user if anything went wrong.  If we thought it was useful to have information about successfully inserted monomers, could that be a separate list?

ricrogz:
> Is it worth logging the successes? If we skipped those log entries, it would let the calling code determine whether the operation was completely successful by just checking .empty() on the returned vector. That would also make it easier for the calling code to present only the failure messages to the user if anything went wrong. If we thought it was useful to have information about successfully inserted monomers, could that be a separate list?

The suggested return for the updated API is a JSON that includes the successful inserts (see https://schrodinger.atlassian.net/browse/SHARED-12481?focusedCommentId=4232736). Given that JSON is impractical to work with in C++, I opted for a vector with the individual outcomes, which has the advantage of presenting the result of each monomer insertion in the same order as they appear in the input JSON data.

KevKeating:
> > Is it worth logging the successes? If we skipped those log entries, it would let the calling code determine whether the operation was completely successful by just checking .empty() on the returned vector. That would also make it easier for the calling code to present only the failure messages to the user if anything went wrong. If we thought it was useful to have information about successfully inserted monomers, could that be a separate list?
> 
> The suggested return for the updated API is a JSON that includes the successful inserts (see https://schrodinger.atlassian.net/browse/SHARED-12481?focusedCommentId=4232736). Given that JSON is impractical to work with in C++, I opted for a vector with the individual outcomes, which has the advantage of presenting the result of each monomer insertion in the same order as they appear in the input JSON data.

Ahh, okay.  In that case, the current behavior definitely sounds reasonable.

KevKeating (approved):

src/app/main.cpp
```diff
@@ -102,12 +108,57 @@ void sketcher_load_custom_monomers_from_sql(const std::string& sql)
...
-void sketcher_insert_custom_monomers(const std::string& json)
+static MonomerDefsInsertionResult
+insert_custom_monomers(const std::string& json)
 {
     auto& db = schrodinger::rdkit_extensions::MonomerDatabase::instance();
-    db.insertMonomersFromJson(json);
+    return db.insertMonomersFromJson(json);
+}
+
+#ifdef __EMSCRIPTEN__
```
cdvonbargen, line 102: perhaps a dumb question -- are these APIs needed for *non* WASM from main.cpp? I think they are only here for the WASM itself, in which case can we just have the WASM ones?

ricrogz: I did not add these interfaces, so I have no idea what the expectations are.

rachelnwalker (approved):
Approving the sketch PR, but one change suggestion on the core suite one.

src/app/main.cpp
cdvonbargen: @KevKeating @JarrettSJohnson -- would it make sense to leave these as wasm-only APIs in this file? It would simplify what Ricardo has here a bit

KevKeating: Yeah, that seems reasonable to me.  The core suite version of the monomer database shouldn't need Sketcher to populate it.

PR #380, opened 2026-06-25
SKETCH-2797: Fixing highlighting path generation

KevKeating:
- Linked Case: SKETCH-2797

### Description

This reverts the AbstractHighlightingItem changes from https://github.com/schrodinger/mmshare/pull/5798 (and adds an explanatory comment), since those changes are causing selection paths to be generated incorrectly  when there's an overlap in the path, most notably for nucleic acids:
image
Here's the same image after this revert:
image

@matvey83, I have vague memories of you saying that the measurable speedup in SKETCH-2604 was from the signal disconnection and not the path generation.  Is that right?  The slowdown in that case happened as soon as the user starting selecting anything, right?  It didn't kick in only when the selection path was complicated?  I just wanted to sanity check that we don't need to worry about re-introducing any performance issues here.  I wouldn't expect these calculations to take any observable amount of time even with a fairly sizable selection.

### Testing Done

Generated the above images.

### DIFF
--- a/src/schrodinger/sketcher/molviewer/abstract_highlighting_item.cpp
+++ b/src/schrodinger/sketcher/molviewer/abstract_highlighting_item.cpp
@@ -60,17 +60,18 @@ QPainterPath AbstractHighlightingItem::buildHighlightingPathForItems(
     const QList<const QGraphicsItem*>& items) const
 {
     QPainterPath path;
-    // Set fill rule to ensure overlapping areas don't create "holes"
-    path.setFillRule(Qt::WindingFill);
-
     for (auto item : items) {
         if (auto* molviewer_item =
                 dynamic_cast<const AbstractGraphicsItem*>(item)) {
             QPainterPath local_path = getPathForItem(molviewer_item);
-            path.addPath(molviewer_item->mapToScene(local_path));
+            // We can't use addPath here, since that will result in the path
+            // being "hollowed out" where the path for two monomers intersects.
+            // E.g., there will be a hole in the selection highlighting between
+            // a nucleic acid sugar and base when both are selected.
+            path |= molviewer_item->mapToScene(local_path);
         }
     }
-    return path.simplified();
+    return path;
 }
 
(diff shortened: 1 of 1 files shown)


src/schrodinger/sketcher/molviewer/abstract_highlighting_item.cpp
```diff
@@ -60,17 +60,18 @@ QPainterPath AbstractHighlightingItem::buildHighlightingPathForItems(
...
         if (auto* molviewer_item =
                 dynamic_cast<const AbstractGraphicsItem*>(item)) {
             QPainterPath local_path = getPathForItem(molviewer_item);
-            path.addPath(molviewer_item->mapToScene(local_path));
+            // We can't use addPath here, since that will result in the path
+            // being "hollowed out" where the path for two monomers intersects.
+            // E.g., there will be a hole in the selection highlighting between
+            // a nucleic acid sugar and base when both are selected.
+            path |= molviewer_item->mapToScene(local_path);
         }
```
ZontaNicola, line 72: is this a evenOdd fill thing that we can fix using a different rule? I guess not because we still need a clean border line, right?

I'm happy to revert, just double checking that we can't have our cake and eat it too (or as we say here, a have a full barrel and a drunk wife)

ZontaNicola (approved):
looks good to me if we can't get away with a different fill rule

src/schrodinger/sketcher/molviewer/abstract_highlighting_item.cpp
KevKeating: Yeah, that's unfortunately correct.  There are only [two fill rules](https://doc.qt.io/qt-6.5/qt.html#FillRule-enum), and neither one of them is equivalent to just unioning the two paths.  I played around with switching the fill rule here as well as applying a fill rule to the returned path, and nothing had the desired effect other than switching to `|=`.

ethan-schrodinger (approved):

PR #381, opened 2026-07-01
SKETCH-2804: updating Eigen and MSVC paths

KevKeating:
- Linked Case: SKETCH-2804

### Description

The Eigen package was recently updated (BLDMGR-11990, [PR](https://github.com/schrodinger/package-factory_recipes/pull/462)).  The version of Eigen is unchanged, but the new package was generated using the new package-factory-native build process instead of being copied from `lib-build-scripts`.  This new package does, however, contain two minor differences:
- the new package is installed to a different directory inside of `$SCHRODINGER/schrodinger_buildenv_packages`
- the new package no longer contains an `Eigen3Config.cmake`
As a result of these changes, local Sketcher builds (and presumably package factory Sketcher builds) using buildinger now complain that they can't find Eigen.  I've posted a [package factory PR](https://github.com/schrodinger/package-factory_recipes/pull/490) to add in an `Eigen3Config.cmake`, and this PR updates `premake` to point at the new Eigen directory.

Additionally, package factory now contains a new version of Visual Studio for Windows builds (VS 2026 instead of VS 2019).  This package slightly tweaks the path to the compiler, so I've updated that here as well.  

### Testing Done

Successfully built Sketcher on Windows using buildinger with a recent pull of mmshare master after manually adding the new `Eigen3Config.cmake` to my `$SCHRODINGER/schrodinger_buildenv_packages`.


ethan-schrodinger (approved):

cdvonbargen (approved):

premake
```diff
@@ -33,7 +33,7 @@ CMAKE_PREFIX_PATH="$PF_DIR;$ZLIB_DIR;$EIGEN_DIR;$COORDGEN_DIR;$MAEPARSER_DIR;$SQ
 # Configure compiler for Windows
 CXX_COMPILER=""
 if [[ "$OSTYPE" == "msys" ]]; then
-    CXX_COMPILER=$(ls -d $PF_DIR/vs-2019/compiler-*/VC/Tools/MSVC/*/bin/Hostx64/x64/cl.exe)
+    CXX_COMPILER=$(ls -d $PF_DIR/vs-*/VC/Tools/MSVC/*/bin/Hostx64/x64/cl.exe)
```
kartlee: since we source `build_env` above, `which cl.exe` is enough to get the complete path of compiler location.

kartlee (approved):
LGTM, except for the question related to sourcing compiler location.

premake
KevKeating: Good point!  I switched to
```
CXX_COMPILER=$(cygpath `which cl.exe`)
```
here.  `cygpath` switches the variable to a Unix-style path, since CMake gave me errors about an invalid compiler when I passed in the Windows-style path.

PR #383, opened 2026-07-02
SKETCH-2807: Exposing image_generation API via JavaScript for Bunsen

KevKeating:
- Linked Case: SKETCH-2807

### Description

Bunsen wanted a way to draw images with highlighting or labels, so I had Codex write JavaScript wrappings for the get_image_bytes() function from image_generation.h.  There's a fair bit of new Emscripten code here, since we need some way to populate a RenderOptions object entirely from JavaScript.

### Testing Done

Codex wrote new Playwright tests and I've confirmed that they pass.  I also ran the WASM Sketcher version locally and generated an image from the JavaScript console by calling
```js
Module.get_image_bytes('CCCCC', 'PNG', {background_color: '#999999', rdatom_index_to_halo_color: {0: '#FF0000', 3: '#0000FF'}})
```
which returned a JavaScript `uint8array` containing the bytes of:
cccccc2


test/wasm/wasm_api.test.js
```diff
@@ -136,6 +136,52 @@ test.describe('WASM Sketcher API', () => {
     });
   });
 
+  ['SVG', 'PNG'].forEach((imageFormat) => {
```
ethan-schrodinger, line 139: Nit: is there a reason we're using these strings and not the `ImageFormat` enums?

test/wasm/wasm_api.test.js
```diff
@@ -136,6 +136,52 @@ test.describe('WASM Sketcher API', () => {
...
+        show_symbol_for_H_isotopes: false,
+        carbon_labels: Module.CarbonLabels.ALL,
+        color_scheme: Module.ColorScheme.DEFAULT,
+      });
+      return Array.from(imageBytes);
+    });
+    const svg = Buffer.from(bytes).toString('utf8');
+
+    expect(svg).toContain('svg width="222px" height="111px"');
+    expect(svg).toContain('AtomZero');
```
ethan-schrodinger, line 186: >Bunsen wanted a way to draw images with highlighting or labels

I guess there's probably no easy way for us to test whether this properly generates highlighting and labels?

ethan-schrodinger (approved):
I had Codex look at this and it flagged a potential (and unrelated) issue with the way that `trim_image` interacts with generating SVGs, and specifically disagreement between the image size and the SVG tag. Is this anything?

>...actual trimmed render size may be roughly 200 x 200, but SVG tag gets rewritten to 200px x 2000px. That can reintroduce visual padding / odd aspect behavior in browser consumers, defeating trim_image.

test/wasm/wasm_api.test.js
KevKeating: I think it's because these strings get interpolated into the test names, so we need an actual string for that.  For the test logic, it looks like they're converted into a real enum via the `Module.ImageFormat[imageFormat]` on line 142.

test/wasm/wasm_api.test.js
KevKeating: Probably not without comparing the entire SVG string or the actual PNG image, both of which seem brittle.

KevKeating (commented):
I just pushed a new commit that moves most of the new code to separate image_generation_from_js.h/.cpp files to help keep main.cpp a little bit cleaner.  There's no new logic in this commit; it's just the code move, some short docstrings in the new header, and a CMake file tweak so that the new .cpp file gets included in the WASM build.

JarrettSJohnson:
Can I ask where this request came from and how'd it would be used in Bunsen? I don't know if Bunsen would accept this as-is for the integrated sketcher because of the backend <-> parent window websocket that it uses.  For when I used the other API for exporting images, it came through as Base64 where it could be properly decoded.

Edit: I could probably make some changes on the Bunsen-side actually to accomidate this.

KevKeating:
> Can I ask where this request came from and how'd it would be used in Bunsen? I don't know if Bunsen would accept this as-is for the integrated sketcher because of the backend <-> parent window websocket that it uses. For when I used the other API for exporting images, it came through as Base64 where it could be properly decoded.
> 
> Edit: I could probably make some changes on the Bunsen-side actually to accomidate this.

The request came from a Bunsen Day interview.  Bunsen asked me if there was any way for it to generate images with highlighting or labels, and I said that there was but that it wasn't exposed to JavaScript.  I asked Bunsen if it would be useful to add JavaScript wrappings for this functionality, and it said that it would be useful if it wanted to point out a particular part of a molecule to the user, since the only other option would be trying to describe things using text.  I've included Bunsen's full answer on the [Jira case](https://schrodinger.atlassian.net/browse/SKETCH-2807) if you wanted more details.

I used the uint8 array as the return type since it seemed like the most direct JavaScript analog to a QByteArray, but I have no objections to switching this to Base64 encoding if you think that would be more helpful.  I'm pretty sure it would be trivial to do on the C++ side of things; just modify `qbyte_array_to_uint8_array` to instead return `bytes.toBase64().toStdString()` (and presumably change the name of the function).

KevKeating:
Also, I forgot to mention that this wouldn't involve an integrated Sketcher window at all.  Bunsen would just pass in a molecule (e.g. SMILES string or molblock) and info on what atoms/bonds it wants labeled or highlighted, and it would get back the bytes of an image.  Behind the scenes, a new temporary `Scene` object would be created and populated without an associated `SketcherWidget`.

ethan-schrodinger (approved):

JarrettSJohnson (approved):
Ack'd

Lemme know when this is merged so i can test the API with bunsen. I think I can do the base64 part on their side before it reaches the websocket

CMakeLists.txt
```diff
@@ -234,7 +234,8 @@ enable_qt_moc(${SKETCHER_TARGET})
 
 # Standalone executable
 set(APP_TARGET schrodinger_sketcher)
-set(APP_SOURCES src/app/main.cpp src/app/crash_handler.cpp)
+set(APP_SOURCES src/app/main.cpp src/app/crash_handler.cpp
```
cdvonbargen: should this glob everything in src/app?

src/app/main.cpp
cdvonbargen, line 65: should we remove this API (doesnt have to be part of this review)

cdvonbargen (approved):

KevKeating:
> I had Codex look at this and it flagged a potential (and unrelated) issue with the way that `trim_image` interacts with generating SVGs, and specifically disagreement between the image size and the SVG tag. Is this anything?
> 
> > ...actual trimmed render size may be roughly 200 x 200, but SVG tag gets rewritten to 200px x 2000px. That can reintroduce visual padding / odd aspect behavior in browser consumers, defeating trim_image.

That's a good point.  I just filed SKETCH-2816 for that.  Should just be a matter of passing the trimmed size into `post_process_svg()'.

src/app/main.cpp
KevKeating: I think it's still worth having this so that it's possible to generate an image of the current Sketcher workspace.  There's no way for JavaScript to fetch the current Sketcher display settings as a RenderOptions object, so there's no way to achieve the results of this function by using the new functions.

CMakeLists.txt
KevKeating: Yeah, that's smart.  I'll change this.

KevKeating (commented):
I've updated `get_image_bytes` to return a Base64-encoded string.  That matches the return type of `sketcher_export_image` and should also save Jarrett a tiny bit of work.  I'll update the diff here as soon as the sanity check run finishes on my fork.

PR #385, opened 2026-07-08
SKETCH-2812: Improving performance of highlighting path merging

KevKeating:
- Linked Case: SKETCH-2812

### Description

My fix for SKETCH-2797 appears to have had a larger impact on performance than I expected based on the results from [this performance test](https://stu.schrodinger.com/performance/test/1016#metric7674-average-image-generation-with-highlighting-smiles-6-molecules-seconds).  I attempted to see the changes from that case had reintroduced the Maestro 2D overlay lag, but I wasn't able to trigger any lag even after reverting all of Matvey's speed ups.  This PR, however, appears to almost completely restore the old performance test behavior, at least according to local runs.  That means that it should hopefully fix any Maestro 2D overlay issues that may or may not have been introduced.

This change simply tweaks how we calculate the union of a bunch of QPainterPaths, using an approach that Codex came up with.  Basically, instead of unioning each path one-by-one onto an increasingly complex accumulated path, we instead use divide-and-conquer to merge the paths by recursively unioning the first half of the path list with the second half of the path list.  This keeps the intermediate paths as small as possible, which speeds up each individual union operation.

I was somewhat skeptical that this would have a substantial impact on performance, but it appears to have almost completely eliminated the slow down we got from SKETCH-2797.  We'll get a better idea of this once the change is merged into mmshare and we get official performance test results, but I'm cautiously optimistic that this will solve any potential issues here.

### Testing Done

Ran the performance tests locally, and confirmed that highlighting paths look correct in the Windows desktop app.

### DIFF
--- a/src/schrodinger/sketcher/molviewer/abstract_highlighting_item.cpp
+++ b/src/schrodinger/sketcher/molviewer/abstract_highlighting_item.cpp
@@ -1,6 +1,7 @@
 #include "schrodinger/sketcher/molviewer/abstract_highlighting_item.h"
 
 #include <functional>
+#include <vector>
 
 #include <QList>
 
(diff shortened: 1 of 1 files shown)


src/schrodinger/sketcher/molviewer/abstract_highlighting_item.cpp
```diff
@@ -56,22 +57,41 @@ QList<const QGraphicsItem*> AbstractHighlightingItem::updateItemsToHighlight(
     return items;
 }
 
+/**
+ * Calculate the union of all specified paths. This function uses
+ * divide-and-conquer to keep the intermediate paths as small as possible, which
+ * dramatically improves performance over a naive for loop implementation.
```
ethan-schrodinger, line 63: Do you have a sense for why keeping the intermediate paths small should improve performance so significantly?

ethan-schrodinger (approved):

rachelnwalker (approved):
Thank you!!!!

src/schrodinger/sketcher/molviewer/abstract_highlighting_item.cpp
KevKeating: Not really.  I've taken a look at the source code for `QPainterPath::united`, but it's quite difficult to follow without spending a real amount of time on it.  (I did find a comment of ```// FIXME: this can be optimized by using binary search``` that's git blamed to the first commit of the repo from 15 years ago, which seems suspiciously relevant to our current problem.)

PR #388, opened 2026-07-13
SKETCH-2815: Publish React sketcher component as @schrodinger/sketcher/react

cdvonbargen:
- Linked Case: SKETCH-2815

### Description

The React binding for the Maestro Sketcher lives in a separate repo, `rc-maestro-sketcher`, which wraps this project's WebAssembly build and re-publishes it. That split forces the component's `RepresentationFormat` enum to be a hand-maintained mirror of the C++ `Format` enum and relies on a nightly cron to chase WASM versions across the repo boundary.

This change makes `sketcher` the single source of truth by publishing the React component from `@schrodinger/sketcher` itself, as a `./react` subpath export alongside the existing WASM assets. A small Rollup/TypeScript build compiles the component into `dist/react` during the existing npm publish flow, with the iframe cache-bust hash derived from the assembled `wasm_shell.html` at build time.

Once published, the two consumers (LiveDesign and ACAS) will import `@schrodinger/sketcher/react` directly and `rc-maestro-sketcher` will be retired.

### Testing Done

Built the package locally and confirmed `dist/react` is produced with `__WASM_HASH__` replaced by the md5 of `wasm_shell.html`; verified `npm pack` includes both `dist/react` and the WASM assets, and that a scratch install resolves `@schrodinger/sketcher/react` with type declarations.

### DIFF
--- a/wasm/package/package.json
+++ b/wasm/package/package.json
@@ -34,14 +34,14 @@
         "build": "rollup -c"
     },
     "peerDependencies": {
-        "react": "^16.8.0"
+        "react": "^18.0.0"
     },
     "devDependencies": {
         "@rollup/plugin-commonjs": "^20.0.0",
         "@rollup/plugin-node-resolve": "^13.0.4",
         "@rollup/plugin-replace": "^5.0.0",
         "@rollup/plugin-typescript": "^8.5.0",
-        "@types/react": "^16.14.0",
+        "@types/react": "^18.0.0",
         "md5-file": "^5.0.0",
         "rollup": "^2.56.3",
         "rollup-plugin-peer-deps-external": "^2.2.4",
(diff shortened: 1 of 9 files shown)


wasm/package/package.json
```diff
@@ -10,7 +10,43 @@
...
+    },
     "files": [
         "dist"
-    ]
+    ],
+    "scripts": {
+        "build": "rollup -c"
+    },
+    "peerDependencies": {
+        "react": "^16.8.0"
```
juzerzarif: Bump this to react@18 and use the jsx-runtime transform (`jsx: react-jsx` in tsconfig.json). LD is already at React 18 and the new transform is better.

juzerzarif (approved):

KevKeating (approved):

PR #389, opened 2026-07-17
Adding CMake dependencies for *.qrc files

KevKeating:
- Linked Case: <issue>

### Description

CMake doesn't understand *.qrc files well enough to rebuild them when a referenced icon, font, etc. changes.  This led to some mild annoyance when I was tweaking the monomeric hydrogen bond icon in SKETCH-2780, so I had Codex add a function to our `CMakeLists.txt`.  The new function parses filenames out of the `<file>...</file>` blocks and adds them as explicit dependencies, so CMake will take those files into account when deciding whether to rebuild the *.qrc file.  This definitely wouldn't be sufficient for _any_ *.qrc file (e.g., it doesn't understand `<qresource prefix="all/files/relative/to/this">`), but it's good enough for our *.qrc files.

### Testing Done

Built Sketcher locally.


ethan-schrodinger (approved):
Nice.

cdvonbargen (approved):

PR #395, opened 2026-07-24
Fix .maezst files being written uncompressed

codesbymustafa:
### Description

`get_compression_type_from_ext()` (in `src/schrodinger/rdkit_extensions/file_format.cpp`) determines a file's compression from its extension. It matched gzip using the bare suffix `"gz"`, so glued extensions like `.maegz` and `.sdfgz` are correctly detected - but it matched zstd using `".zst"`. As a result the supported glued MAESTRO extension `.maezst` did not end in `.zst`, fell through to `CompressionType::UNKNOWN`, and was written **uncompressed under a compressed filename**, silently mislabeling the output file.

The write path (`maybe_compressed_ostream`) relies on this function to decide whether to apply a compressor, so the mislabeling is silent on write. The read path sniffs the file's magic bytes rather than the extension, so it doesn't surface the problem either - the file just isn't the format its name claims.

This change matches zstd with the bare `"zst"` suffix, mirroring the existing gzip handling, so `.maezst` is correctly detected as `ZSTD`.

### Testing Done

Added a `.maezst` -> `ZSTD` case to `TestGetCompressionType` in `test/schrodinger/rdkit_extensions/test_file_format.cpp`, along with a `methane.maezst` fixture.

The test exercises both detection paths: 
`get_compression_type()` (content/magic-byte sniff) and 
`get_compression_type_from_ext()` (extension-based).

Existing cases (`.maegz`, `.mae.zst`, `.sdfgz`, `.smigz`, etc.) are unchanged and still pass.


codesbymustafa:
## Repro / summary for reviewers

**Root cause** — the two branches in `get_compression_type_from_ext()` aren't symmetric:

```cpp
if (boost::algorithm::iends_with(filename.string(), "gz")) {          // bare "gz"  → catches .maegz
    return CompressionType::GZIP;
} else if (boost::algorithm::iends_with(filename.string(), ".zst")) { // ".zst"     → misses .maezst
    return CompressionType::ZSTD;
}
```

`.maezst` is a supported MAESTRO extension (it's in `MOL_FORMAT_EXTS` for `Format::MAESTRO`), but it doesn't end in `.zst`, so it falls through to `UNKNOWN`. On the **write** path that means no compressor is applied and the file is written as plain text under a `.maezst` name.

**Observable behavior** — write any structure to a `*.maezst` file and inspect the first bytes:

_Before this change:_

```console
$ xxd -l 8 out.maezst
00000000: 7b0a 2020 2073 5f6d 5f6d              {.   s_m_m     # plain MAESTRO text — not compressed
```

_After this change:_

```console
$ xxd -l 8 out.maezst
00000000: 28b5 2ffd ....                                       # zstd magic 28 b5 2f fd
```

The **read** path isn't affected because it sniffs magic bytes rather than the extension — which is also why the mismatch is silent: the file simply isn't the format its name claims.

**Test** — the added `.maezst → ZSTD` case in `TestGetCompressionType` covers both detection paths (`get_compression_type()` content sniff and `get_compression_type_from_ext()` extension check), backed by a real zstd-compressed `methane.maezst` fixture.

PR #396, opened 2026-07-24
SKETCH-2700: Tests now run with --platform offscreen

KevKeating:
- Linked Case: SKETCH-2700

### Description

The unit tests now run with --platform offscreen, so they shouldn't cause windows to pop up on the screen while they're running.  There are two related changes here that were required to get this working:
- the tests load the Arimo font.  Apparently, the tests had been running with a fallback font.  Previously, this didn't matter too much, but the offscreen platform doesn't bother to read in fonts from the operating system.  (Presumably since there's no chance of anyone actually seeing the font.)  As a result, the tests were being run with a weird enough font that AtomItem's text eliding was returning unexpected results.
- In static builds (i.e. the OSS builds used by the GitHub runner, but not the $SCHRODINGER builds), the tests executables are now statically linked against Qt's offscreen platform.  Otherwise, all of the GitHub runners would crash when they tried to run the tests.

### Testing Done

The tests pass with a local $SCHRODINGER build, and all GitHub runners passed when I ran them on my fork.

### DIFF
--- a/CMakeLists.txt
+++ b/CMakeLists.txt
@@ -399,6 +399,11 @@ if(ENABLE_TESTING)
     target_include_directories(${TEST_TARGET}
                                PRIVATE ${PROJECT_SOURCE_DIR}/test)
     target_link_libraries(${TEST_TARGET} PRIVATE ${TEST_DEPENDENCIES})
+    if(NOT QT6_IS_SHARED_LIBS_BUILD)
+      # we use the offscreen platform to make sure that the tests don't pop up
+      # windows on the screen while they're running
+      qt_import_plugins(${TEST_TARGET} INCLUDE Qt6::QOffscreenIntegrationPlugin)
+    endif()
     if(MSVC)
       # ignore discarded [[nodiscard]] values for the tests
       target_compile_options(${TEST_TARGET} PRIVATE -wd4834)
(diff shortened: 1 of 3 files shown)


ethan-schrodinger (approved):
>Apparently, the tests had been running with a fallback font. Previously, this didn't matter too much, but the offscreen platform doesn't bother to read in fonts from the operating system. (Presumably since there's no chance of anyone actually seeing the font.) As a result, the tests were being run with a weird enough font that AtomItem's text eliding was returning unexpected results.

Fascinating.

CMakeLists.txt
```diff
@@ -399,6 +399,11 @@ if(ENABLE_TESTING)
     target_include_directories(${TEST_TARGET}
                                PRIVATE ${PROJECT_SOURCE_DIR}/test)
     target_link_libraries(${TEST_TARGET} PRIVATE ${TEST_DEPENDENCIES})
+    if(NOT QT6_IS_SHARED_LIBS_BUILD)
```
cdvonbargen, line 402: Is it worth a spin out case to do this for the static libs too? Or is that silly

cdvonbargen (approved):

CMakeLists.txt
KevKeating: The offscreen stuff currently works for both the static and shared builds.  The static build needs the offscreen platform included at compile time, which is what this block does.  The shared builds have the offscreen plugin installed by default and automatically load it when needed.

PR #397, opened 2026-07-27
SKETCH-2752, SKETCH-2739: Fixing keyboard shortcuts

KevKeating:
- Linked Case: SKETCH-2752, SKETCH-2739

### Description

Both of these cases had the same underlying issue: the SketcherWidget wasn't fully updated to handle the SketcherModel changes required for non-native monomers.  That change added two new model keys: `ModelKey::AMINO_ACID_SYMBOL` and `ModelKey::NUCLEIC_ACID_SYMBOL` (which are strings) to go along with the existing `ModelKey::AMINO_ACID_TOOL` and `ModelKey::NUCLEIC_ACID_TOOL`  keys (which are enums).  `SketcherWidget::keyPressEvent` only updated the `_TOOL` keys, though, not the `_SYMBOL` keys.

While I was fixing this, I also made one additional change to how monomeric keyboard shortcuts are handled: if there's a selection containing only amino acids or only nucleic acids, then we use the selection to determine whether the keyboard shortcut should be interpreted as an amino acid shortcut or a nucleic acid shortcut instead of using the current side bar contents.  That's almost certainly what the user intended, especially since the alternative is guaranteed to be a no-op.

### Testing Done

Tested on the Windows desktop app, and added a bunch of new unit tests

### DIFF
--- a/src/schrodinger/sketcher/sketcher_widget.cpp
+++ b/src/schrodinger/sketcher/sketcher_widget.cpp
@@ -1582,11 +1582,9 @@ void SketcherWidget::handleNucleicAcidKeyboardShortcuts(
          QVariant::fromValue(MonomerToolType::NUCLEIC_ACID)},
     };
 
-    // if there's a selection (i.e. has_targets is true), then the keypress is
-    // going to mutate monomers so we don't care what current_tool is (since the
-    // real current tool is the selection tool, not a nucleic acid tool)
     if (!has_targets && (current_tool == NucleicAcidTool::RNA_NUCLEOTIDE ||
                          current_tool == NucleicAcidTool::DNA_NUCLEOTIDE)) {
+        // a full nucleotide tool is active
         if (BASE_KEYS.contains(key)) {
             // switch the base of the current full nucleotide tool
             auto base = std::get<StdNucleobase>(BASE_KEYS.at(key));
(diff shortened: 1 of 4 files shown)


src/schrodinger/sketcher/sketcher_widget.cpp
```diff
@@ -1546,68 +1575,82 @@ void SketcherWidget::handleNucleicAcidKeyboardShortcuts(
...
+    bool has_targets = !targets.atoms.empty();
+    std::unordered_map<ModelKey, QVariant> kv_pairs = {
+        {ModelKey::DRAW_TOOL, QVariant::fromValue(DrawTool::MONOMER)},
+        {ModelKey::MONOMER_TOOL_TYPE,
+         QVariant::fromValue(MonomerToolType::NUCLEIC_ACID)},
+    };
+
+    // if there's a selection (i.e. has_targets is true), then the keypress is
+    // going to mutate monomers so we don't care what current_tool is (since the
+    // real current tool is the selection tool, not a nucleic acid tool)
```
ethan-schrodinger: It looks like this comment is concerned with the case where `has_targets = true`, but we don't actually address that case until l1625. Should this comment be moved there? Would it be helpful to insert a comment here that's more germane to the more proximate cases (`has_targets = false`)?

ethan-schrodinger (approved):

src/schrodinger/sketcher/sketcher_widget.cpp
KevKeating: I was trying to explain why `!has_targets` was included in the condition, but I think this comment winds up being more confusing than helpful.  I got rid of it and added some brief explanatory comments about the first two blocks.

PR #402, opened 2026-08-06
SKETCH-2832: SMILES monomers now show atomistic image in tool tip

KevKeating:
- Linked Case: SKETCH-2832

### Description

SMILES monomers now have a tool tip that shows an atomistic image of the SMILES molecule:
image
This is implemented using the standard Qt tooltips, which can handle images if they're passed as HTML containing a base64-encoded image.  Unfortunately, this means that the image rendering is a bit aliased, as you can see in the screen shot above.  The only way to fix this would be to manually implement our own tool-tip-like behavior using mouse events.  I've done this for Python panels before I learned about the base64-encoding trick, but it takes a bit of work.  It doesn't seem worth worrying about until Laura complains about it.

Also, as a result of this change, monomer graphics items now take `AtomDisplaySettings` and `BondDisplaySettings` objects so that the SMILES molecule can be rendered using the current atomistic settings.  This change required updating a bunch of signatures of constructors and other functions so that we could pass settings from the `Scene` to the graphics items and scene tools (which instantiate graphics items for hints when hovering over monomers).  That's the main reason why this diff touches so many files.  The only new logic is in abstract_monomer_item.cpp; everything else is just small tweaks to pass these settings around.

While I was working on this case, I filed SKETCH-2842 to change the label for SMILES monomers from `***` to a small thumbnail of the tool tip image.  What do people think of that idea?  The image won't be large enough to make out all of the details, but it should at least give an idea of what the atomistic structure is, and it seems a lot more informative than just `***`.  It shouldn't be difficult to implement now that the monomeric graphics items receive the atom and bond settings.  We'd probably want to make the bond lines a little thicker so they're visible at the small size, similar to what we do for cursor hints.

### Testing Done

Tested on the Windows desktop add and added a new unit test that sanity checks tool tips.

### DIFF
--- a/src/schrodinger/sketcher/molviewer/abstract_monomer_item.cpp
+++ b/src/schrodinger/sketcher/molviewer/abstract_monomer_item.cpp
@@ -2,13 +2,19 @@
 
 #include <cmath>
 
+#include <QBuffer>
+#include <QGraphicsScene>
 #include <QPainter>
+#include <QSvgGenerator>
 
 #include <rdkit/GraphMol/Atom.h>
 #include <rdkit/GraphMol/MonomerInfo.h>
 
 #include "schrodinger/sketcher/image_constants.h"
 #include "schrodinger/sketcher/molviewer/coord_utils.h"
+#include "schrodinger/sketcher/molviewer/scene_utils.h"
+#include "schrodinger/sketcher/rdkit/mol_update.h"
+#include "schrodinger/rdkit_extensions/convert.h"
 #include "schrodinger/rdkit_extensions/helm.h"
 #include "schrodinger/rdkit_extensions/monomer_database.h"
 #include "schrodinger/rdkit_extensions/monomer_mol.h"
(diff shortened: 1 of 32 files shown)


ethan-schrodinger (approved):
>I've done this for Python panels before I learned about the base64-encoding trick, but it takes a bit of work. It doesn't seem worth worrying about until Laura complains about it.

It would probably be polite of us to at least spin off a ticket to address this so that Laura doesn't have to.
>While I was working on this case, I filed SKETCH-2842 to change the label for SMILES monomers from *** to a small thumbnail of the tool tip image. What do people think of that idea? The image won't be large enough to make out all of the details, but it should at least give an idea of what the atomistic structure is, and it seems a lot more informative than just ***.

I don't think this is a bad idea, but we should present it to Laura in such a way that it doesn't look like we're trying to circumvent her spec.

KevKeating:
> > I've done this for Python panels before I learned about the base64-encoding trick, but it takes a bit of work. It doesn't seem worth worrying about until Laura complains about it.
> 
> It would probably be polite of us to at least spin off a ticket to address this so that Laura doesn't have to.

Yeah, that's smart.  I had one idea that I wanted to test to fix this: generate the image at a higher-resolution than we're going to display, and then specify the desired dimensions in the ` 
> > While I was working on this case, I filed SKETCH-2842 to change the label for SMILES monomers from *** to a small thumbnail of the tool tip image. What do people think of that idea? The image won't be large enough to make out all of the details, but it should at least give an idea of what the atomistic structure is, and it seems a lot more informative than just ***.
> 
> I don't think this is a bad idea, but we should present it to Laura in such a way that it doesn't look like we're trying to circumvent her spec.

Yeah, good idea.  I'll spend a few minutes on this to generate a quick screen shot so Laura can see what I'm talking about and then ping her on the case.

KevKeating:
> I had one idea that I wanted to test to fix this: generate the image at a higher-resolution than we're going to display, and then specify the desired dimensions in the `image| image| 

This definitely _shouldn't_ work since the images are vector-based SVG, but Qt must use the SVGs stated resolution when rasterizing the image.  Regardless, I'll update the diff and we don't have to worry about a spin-off case.

PR #403, opened 2026-08-10
ensure consistent CW spelling of rings

ZontaNicola:
- Linked Case: CRDGEN-418

### Description

looks like rdkit normalizeRing returns ring spelt out in opposite directions for add and even... not sure why. This flips coordinates in half of the cases so they are all spelt out CW. Note that this is only done when there's a single ring of monomers without any other decoration

PEPTIDE1{C.P.M.H.V.I.K.C}|CHEM1{[mDBX]}$PEPTIDE1,CHEM1,1:R3-1:R1|PEPTIDE1,CHEM1,8:R3-1:R2$$$V2.0

PEPTIDE1{C.P.M.M.H.V.I.K.C}|CHEM1{[mDBX]}$PEPTIDE1,CHEM1,1:R3-1:R1|PEPTIDE1,CHEM1,9:R3-1:R2$$$V2.0

Screenshot 2026-08-10 at 12 32 52

### Testing Done

added a test

### DIFF
--- a/src/schrodinger/rdkit_extensions/helm/monomer_coordgen.cpp
+++ b/src/schrodinger/rdkit_extensions/helm/monomer_coordgen.cpp
@@ -511,6 +511,29 @@ generate_coordinates_for_cycles(RDKit::ROMol& polymer,
     params.forceRDKit = true;
     params.useRingTemplates = false;
     RDDepict::compute2DCoords(polymer, params);
+
+    // Ensure consistent ring direction (CW) for single-ring polymers by
+    // checking if the ring atoms are laid out in ascending backbone order.
+    // RDKit's normalizeRing can produce different orderings for different ring
+    // sizes. If ascending (CCW), mirror the y-coordinates to flip to CW.
+    // Only apply this fix for single rings where ALL atoms are in the ring,
+    // to avoid disrupting layouts with external attachments.
+    auto& conformer = polymer.getConformer();
+    const auto& rings = polymer.getRingInfo()->atomRings();
+    if (rings.size() == 1 && rings[0].size() == polymer.getNumAtoms()) {
+        const auto& ring = rings[0];
+        if (ring.size() > 2 && ring[1] < ring.back()) {
+            // Ring is in ascending order (CCW) - mirror y-coordinates around
+            // centroid to make it CW
+            auto centroid = compute_centroid(polymer, ring);
+            for (auto atom_idx : ring) {
+                auto pos = conformer.getAtomPos(atom_idx);
+                pos.y = 2 * centroid.y - pos.y;
+                conformer.setAtomPos(atom_idx, pos);
+            }
+        }
+    }
+
     orient_ring_system(polymer);
 
     // finalize coordinates of rings and of their immediate neighbors.
--- a/test/schrodinger/rdkit_extensions/test_monomer_coordgen.cpp
+++ b/test/schrodinger/rdkit_extensions/test_monomer_coordgen.cpp
@@ -743,4 +743,52 @@ BOOST_AUTO_TEST_CASE(TestPositionInCoilsScore)
     }
 }
 
+BOOST_AUTO_TEST_CASE(CyclicRingsHaveConsistentDirection)
+{
+    // Test that cyclic rings are always drawn in the same direction (CCW)
+    // regardless of odd/even ring size. RDKit's normalizeRing can produce
+    // different atom orderings for different ring sizes, but our fix ensures
+    // consistent visual direction.
+
+    // 9-atom ring (odd)
+    std::string helm_odd =
+        "PEPTIDE1{C.P.M.H.V.I.K.C}|CHEM1{[mDBX]}$PEPTIDE1,CHEM1,1:R3-1:R1|"
+        "PEPTIDE1,CHEM1,8:R3-1:R2$$$V2.0";
+    // 10-atom ring (even)
+    std::string helm_even =
+        "PEPTIDE1{C.P.M.M.H.V.I.K.C}|CHEM1{[mDBX]}$PEPTIDE1,CHEM1,1:R3-1:R1|"
+        "PEPTIDE1,CHEM1,9:R3-1:R2$$$V2.0";
+
+    auto mol_odd = helm::helm_to_rdkit(helm_odd);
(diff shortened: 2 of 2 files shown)


ZontaNicola:
I think this is necessary for now, but maybe we can address this in rdkit directly? I don't see why it should mess up the order of atom rings

rachelnwalker (commented):
@ricrogz - didn't you fix something like this in RDKit? https://github.com/rdkit/rdkit/commit/9e301c15d6c2f5cab848ffaa788300505d69ff94?

ricrogz:
> @ricrogz - didn't you fix something like this in RDKit? [rdkit/rdkit@9e301c1](https://github.com/rdkit/rdkit/commit/9e301c15d6c2f5cab848ffaa788300505d69ff94)?

It wasn't a fix. Before I replaced the ring finding algoritm (to which we haven't caught up yet, it's coming in the fall release), I made RDKit store `atomRing` indexes so that the first index is the lowest one, and then enumerating the ring in the direction of the lowest indexed neighbor. This was to make tests as stable as possible (i.e. that test results didn't change depending how rings where found/enumerated).

This had nothing to do with coordinates, though. It wasn't meant to make rings CW or CCW.

test/schrodinger/rdkit_extensions/test_monomer_coordgen.cpp
```diff
@@ -743,4 +743,52 @@ BOOST_AUTO_TEST_CASE(TestPositionInCoilsScore)
     }
 }
 
+BOOST_AUTO_TEST_CASE(CyclicRingsHaveConsistentDirection)
+{
+    // Test that cyclic rings are always drawn in the same direction (CCW)
+    // regardless of odd/even ring size. RDKit's normalizeRing can produce
+    // different atom orderings for different ring sizes, but our fix ensures
+    // consistent visual direction.
```
KevKeating, line 751: You could switch this to a test docstring if you wanted since it applies to the whole test.  Either way is fine by me, though.

src/schrodinger/rdkit_extensions/helm/monomer_coordgen.cpp
```diff
@@ -511,6 +511,29 @@ generate_coordinates_for_cycles(RDKit::ROMol& polymer,
...
+            // Ring is in ascending order (CCW) - mirror y-coordinates around
+            // centroid to make it CW
+            auto centroid = compute_centroid(polymer, ring);
+            for (auto atom_idx : ring) {
+                auto pos = conformer.getAtomPos(atom_idx);
+                pos.y = 2 * centroid.y - pos.y;
+                conformer.setAtomPos(atom_idx, pos);
+            }
+        }
+    }
```
KevKeating, line 535: You could pull this out into a separate static function and switch the comment to a docstring.  That would presumably make it slightly easier to remove once it's no longer needed.

KevKeating (commented):
> > @ricrogz - didn't you fix something like this in RDKit? [rdkit/rdkit@9e301c1](https://github.com/rdkit/rdkit/commit/9e301c15d6c2f5cab848ffaa788300505d69ff94)?
> 
> It wasn't a fix. Before I replaced the ring finding algoritm (to which we haven't caught up yet, it's coming in the fall release), I made RDKit store `atomRing` indexes so that the first index is the lowest one, and then enumerating the ring in the direction of the lowest indexed neighbor. This was to make tests as stable as possible (i.e. that test results didn't change depending how rings where found/enumerated).
> 
> This had nothing to do with coordinates, though. It wasn't meant to make rings CW or CCW.

I think 9e301c1 happens to fix this issue as a side effect.  Codex claims that the source of the underlying problem here is the inconsistent atom order for rings, since RDKit's `embedRing` always places the ring counterclockwise based on the atom index order that its given.  Codex also claims that the current code in RDKit's `master` branch shouldn't have this issue.  I made a Sketcher build using an RDKit commit from today (`7e5eeab06e80628c3ace5d1cee24129a6fc923dc`), and ring direction appears to be consistent, at least on the Windows desktop app.  Pasting all of these HELM strings
```
PEPTIDE1{C.P.M.H.V.I.K.C}|CHEM1{[mDBX]}$PEPTIDE1,CHEM1,1:R3-1:R1|PEPTIDE1,CHEM1,8:R3-1:R2$$$V2.0
PEPTIDE1{C.P.M.H.V.I.K.C.C}|CHEM1{[mDBX]}$PEPTIDE1,CHEM1,1:R3-1:R1|PEPTIDE1,CHEM1,9:R3-1:R2$$$V2.0
PEPTIDE1{C.P.M.H.V.I.K.C.C.C}|CHEM1{[mDBX]}$PEPTIDE1,CHEM1,1:R3-1:R1|PEPTIDE1,CHEM1,10:R3-1:R2$$$V2.0
PEPTIDE1{C.P.M.H.V.I.K.C.C.C.C}|CHEM1{[mDBX]}$PEPTIDE1,CHEM1,1:R3-1:R1|PEPTIDE1,CHEM1,11:R3-1:R2$$$V2.0
```
gives me
image
(The build is [here](https://github.com/KevKeating/sketcher/actions/runs/31520005176) if anyone wants to play with it.  The test failures are from coordgen tests on cyclic peptides, presumably as a result of this same change.)

I'm guessing we still want this PR's fix until we switch to the next release of RDKit?  Does that seem reasonable?  Maybe with a comment that we can get rid of it once we upgrade to RDKit's fall of 2026 release?  Regardless of what we do with the fix, the unit test is definitely helpful; we should keep that either way.

ZontaNicola:
> I think 9e301c1 happens to fix this issue as a side effect. Codex claims that the source of the underlying problem here is the inconsistent atom order for rings, since RDKit's `embedRing` always places the ring counterclockwise based on the atom index order that its given. Codex also claims that the current code in RDKit's `master` branch shouldn't have this issue. I made a Sketcher build using an RDKit commit from today (`7e5eeab06e80628c3ace5d1cee24129a6fc923dc`), and ring direction appears to be consistent, at least on the Windows desktop app. Pasting all of these HELM strings

If this is fixed in the next RDKit maybe we can live without this patch? @rachelnwalker ? Only thing is I think that most people would read a ring clockwise, and from Kevin's example it looks like we consistently doing CCW? we either keep this patch or just flip a sign and wait for the new rdkit. I don't mind either way

PR #405, opened 2026-08-11
Cherry-pick four recent commits to rdkit_extensions

i-tub:
This applies the following patches from mmshare:

SKETCH-2782: Update residue-number test after chain assignment
SHARED-12108: Support multiple chains in assignChains() (#10872)
BIOLUM-5768: Add MonomerDB check and AP check to standardizer
SHARED-12567: Increment peptide chains based on disconnected residues


test/schrodinger/sketcher/model/test_mol_model.cpp
```diff
@@ -4826,15 +4826,19 @@ BOOST_AUTO_TEST_CASE(test_get_residue_number_for_new_monomer_peptide)
...
                           ap_model_name_for(PeptideAP::N), peptide,
                           ap_model_name_for(PeptideAP::C));
     peptide = model.getMol()->getAtomWithIdx(0);
+    BOOST_TEST(rdkit_extensions::get_residue_number(peptide) == 1u);
+    BOOST_TEST(rdkit_extensions::get_residue_number(
+                   model.getMol()->getAtomWithIdx(1)) == 2u);
     next_peptide_num = get_residue_number_for_new_monomer(
         "C", ChainType::PEPTIDE, ap_model_name_for(PeptideAP::N), peptide);
-    BOOST_TEST(next_peptide_num == 12u);
+    BOOST_TEST(next_peptide_num == 3u);
```
rachelnwalker, line 4841: We should probably make sure this is OK with @KevKeating

rachelnwalker (approved):
Approving but adding @KevKeating as well. I think that the updated behavior is correct, but I want to make sure it won't break anything in sketcher

test/schrodinger/sketcher/model/test_mol_model.cpp
KevKeating: Yeah, that behavior is fine by me.  MolModel doesn't care all that much about the specific numbering.  The test here was mostly to make sure that Sketcher doesn't producing anything that would lead to errors when passed to rdkit_extensions functions.

ethan-schrodinger (approved):

PR #406, opened 2026-08-12
SKETCH-2839: Fixing monomer connection tool icons

KevKeating:
- Linked Case: SKETCH-2839

### Description

The main issue here was that the monomer_tool_widget.ui file had set the icons of the monomeric connection tools be displayed at 16x16 pixels instead of the 30x32 pixels that all of the other buttons use.  As a result, the icons were shrunk down to fit, resulting in very thin lines.  This PR fixes the UI file and replaces the icons with ones that Laura tweaked a bit.  I've also centered the two connection buttons (at Laura's request).  Here's what the side bar looks like now:

image

That screen shot got Laura's seal of approval on the Jira case, so these buttons should be good to go.

(Ignore the fact that GitHub says that the new versions of the icons are 300px by 30px.  GitHub is getting oddly creative with the conversion from SVG pixels to rasterized pixels, but Qt doesn't take the same liberties.)

### Testing Done

Confirmed that the buttons look correct on the Windows desktop version of the app.

### DIFF
--- a/src/schrodinger/sketcher/icons/monomeric_connection.svg
+++ b/src/schrodinger/sketcher/icons/monomeric_connection.svg
@@ -1,29 +1,13 @@
-<?xml version="1.0" encoding="UTF-8" standalone="no"?>
-
-<svg
-   version="1.1"
-   id="Layer_1"
-   x="0px"
-   y="0px"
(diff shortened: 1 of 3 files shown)


ethan-schrodinger (approved):
Great.

PR #407, opened 2026-08-12
SKETCH-2845: Preventing Qt from trying to auto-apply dark mode color scheme on desktop builds

KevKeating:
- Linked Case: SKETCH-2845

### Description

Now that the core suite has upgraded to Qt 6.8, I’m seeing the same behavior from SKETCH-2651 (which was WASM-only) on `$SCHRODINGER` builds of the desktop app.  The result looks considerably less terrible than the result from SKETCH-2651, but I’d hesitate to call it “good.”  This commit just moves the SKETCH-2651 fix out of its `#ifdef __EMSCRIPTEN__` block so that the fix applies to desktop builds as well.

Ethan, since Chris is out this week and this is a fairly trivial change, I'm planning to override the requirement for his approval after you approve this PR.  Let me know if you think it's worth waiting for Chris to take a look at this after R&R before merging.

### Testing Done

Tested on a Windows desktop `$SCHRODINGER` build.  Here's the screen shot before this fix:
image

and here's the screen shot after this fix:
image

### DIFF
--- a/src/app/main.cpp
+++ b/src/app/main.cpp
@@ -311,13 +311,6 @@ EMSCRIPTEN_BINDINGS(sketcher)
 
 void apply_stylesheet(QApplication& app)
 {
-    // In Qt 6.8 and newer, Qt will try to automatically apply a dark mode color
-    // scheme if the system and/or browser is set to dark mode. The result looks
-    // terrible, so switch back to light mode.
-#if QT_VERSION >= QT_VERSION_CHECK(6, 8, 0)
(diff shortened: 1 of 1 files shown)


ethan-schrodinger (approved):
>Ethan, since Chris is out this week and this is a fairly trivial change, I'm planning to override the requirement for his approval after you approve this PR. Let me know if you think it's worth waiting for Chris to take a look at this after R&R before merging.

Seems logical to me. As long as he knows that this happened, I don't think he'll need/want to review it.

PR #408, opened 2026-08-13
make connection line diamonds stem from any side or corner, not  just above or below

ZontaNicola:
- Linked Case: SKETCH-2822

### Description

this was mostly codex doing.
diamonds for side connections were always either top or bottom, this makes them be in any position, rotating so they stay normal to the connection.  I also made the diamond a little smaller to make the layout cleaner.

Here's the structure from the case
Screenshot 2026-08-13 at 11 56 54

and [the full test set](https://zontanicola.github.io/sketcher/galleries/helm-gallery-latest-sketcher.html)

UPDATE: after back and forth with Laura, instead of any direction, we now select either a side or a corner to spawn from. I also introduced a step to avoid bond crossings. The case of intra-polymer disulphides that Kevin mentioned is special cased so that the disulphide doesn't cross the covalent bond

[updated gallery](https://zontanicola.github.io/sketcher/galleries/helm_gallery.html)

### Testing Done

added a test. I also checked the test set to make sure nothing went wrong.
I noticed that there's overlapping for larger labels monomers, but that has nothing to do with this case and we already have a case for it


ethan-schrodinger (commented):
Thanks, @ZontaNicola. Two questions:
1. Which ticket is this associated with?
2. I'm trying to track down where the "rules" for drawing these connection line diamonds are. Can you point me to the right document/page?

ZontaNicola:
1) sorry I had left the angular brackets and the link wasn't showing up!

2) 'm not sure where/if the current behaviour is documented  (@KevKeating might). The behaviour I suggest is a simple line that connects the center of the objects, like any other bond. again asking @KevKeating if this looks reasonable. A full table of how this look is linked in the description of this PR

KevKeating:
I think this was an earlier version of the spec said that diamonds should be either above or below the monomer, but I'm not sure which document that was in.  I don't see anything about it in the current version of the spec, so it's possible that that section got rewritten.  (I also thought I'd filed a case to improve diamond positioning for chains that were laid out vertically, since the current code (i.e. before this PR) is terrible at that) but I don't see that, so maybe I never filed that?)

My main concern here would be for disulfides between two cysteines that are relatively close in sequence.  This is less of a concern than it used to be because of coordgen improvements, but I think it's still worth worrying about.  For example, if a user pastes
```
PEPTIDE1{A.C.A.A.C.A}$PEPTIDE1,PEPTIDE1,2:R3-5:R3$$$V2.0
```
into Sketcher, then we generate:
from HELM
so that the cysteines are laid out vertically.  However, if a user pastes
```
PEPTIDE1{A.C.A.A.C.A}$$$$V2.0
```
into Sketcher and then manually adds the disulfide, then the `main` branch creates
main from manual layout
but this branch shows
new diamonds from manual layout
You can also see similar behavior in the HELM gallery for 
```
PEPTIDE1{C.C}$PEPTIDE1,PEPTIDE1,1:R3-2:R3$$$
```
The main branch shows
main CC
but this branch shows
branch CC

Could we tweak this logic so it takes other bonds into account?  So something like:
  - for the starting monomer, check if the diamond bond is within X° (maybe 10° or so?) of another bond from that monomer.  If it is, place the diamond so it's 90° or 270° offset from the diamond bond's angle (i.e. place the diamond above or below the monomer in the examples above).  If 90° or 270° winds up within X° of another bond from that monomer, prefer the other angle.  Otherwise, maybe prefer the one with the lower Y coordinate, i.e. go above the monomer instead of below, or do something else arbitrary but reproducible
  - Do the same thing for the ending monomer

That would reproduce the `main` behavior in the examples above, while keeping the new behavior (which is a dramatic improvement) in almost all other scenarios.

ZontaNicola:
oh yeah I like this!|

rachelnwalker (commented):
@ZontaNicola - Looking just at the gallery, I have one comment: 946 seems to have a gpa between the end of the connection and the cysteine monomer. I see that on 950 as well

Image

KevKeating:
> @ZontaNicola - Looking just at the gallery, I have one comment: 946 seems to have a gpa between the end of the connection and the cysteine monomer. I see that on 950 as well

It looks like the gap is due to the difference between the corner of the rectangle and the curve of the rounded rectangle.  Presumably, adjusting diagonal coordinates by `AA_ROUNDING_RADIUS` would fix that.

src/schrodinger/sketcher/molviewer/monomer_attachment_point_labels.cpp
```diff
@@ -97,37 +97,37 @@ static void position_ap_label_rect(
...
+        auto unit_x = normal.dx() / normal_length;
+        auto unit_y = normal.dy() / normal_length;
+        auto label_half_extent = qAbs(unit_x) * ap_label_rect.width() / 2 +
+                                 qAbs(unit_y) * ap_label_rect.height() / 2;
+        auto spacing =
+            MONOMER_CONNECTOR_ARROWHEAD_RADIUS +
+            MONOMERIC_ATTACHMENT_POINT_LABEL_ARROWHEAD_SPACING_HORIZONTAL +
+            label_half_extent;
+        normal.setLength(spacing);
+        ap_label_rect.moveCenter(arrowhead_offset + normal.p2());
```
KevKeating, line 132: When labeling a connection, this keeps the labels for the two attachment points on opposite sides of the connection line, right?  That behavior (labels on opposite sides) is fairly important to Laura.  (It looks like it should.  I just wanted to double check.)

src/schrodinger/sketcher/rdkit/monomeric.cpp
```diff
@@ -220,24 +232,452 @@ does_connector_have_arrowheads(const RDKit::Bond* bond,
     }
 }
 
+static std::pair<Direction, Direction>
+get_nearest_cardinal_directions(const QPointF& relative_pos)
```
KevKeating: It looks like only the first value of the returned pair is ever used, so it's probably clearer to only return that value.  That would also make the functionality more closely match the function name.

src/schrodinger/sketcher/rdkit/monomeric.cpp
```diff
@@ -220,24 +232,452 @@ does_connector_have_arrowheads(const RDKit::Bond* bond,
...
+        relative_pos.x() < 0 ? Direction::W : Direction::E;
+    const auto vertical_side =
+        relative_pos.y() > 0 ? Direction::S : Direction::N;
+    if (qAbs(relative_pos.x()) >= qAbs(relative_pos.y())) {
+        return {horizontal_side, vertical_side};
+    }
+    return {vertical_side, horizontal_side};
+}
+
+static QPointF get_direction_unit_vector(const Direction direction)
```
KevKeating, line 248: If you wanted, I think you could implement this using `direction_to_qt_vector` from `unbound_monomeric_attachment_point_item.h`.  (You'd presumably want to move that function someplace more central first.  Maybe molviewer/coord_utils? ).  I think you'd just need to multiple the return value by `COSINE_45_DEGREES` for diagonals.  That would avoid some (very) minor code duplication.

src/schrodinger/sketcher/rdkit/monomeric.cpp
```diff
@@ -220,24 +232,452 @@ does_connector_have_arrowheads(const RDKit::Bond* bond,
...
+static double direction_closeness(const Direction direction,
+                                  const QPointF& relative_pos)
+{
+    const auto unit_vector = get_direction_unit_vector(direction);
+    return QPointF::dotProduct(unit_vector, relative_pos);
+}
+
 /**
- * Return true if coord is above (within a 45 degree cone of) other
+ * Select the four closest forward-facing directions, keeping the nearest side
```
KevKeating: What does "forward-facing" refer to here?  Does that just mean not on the back half of the monomer?

src/schrodinger/sketcher/rdkit/monomeric.cpp
```diff
@@ -220,24 +232,452 @@ does_connector_have_arrowheads(const RDKit::Bond* bond,
...
+    const auto* current_bond =
+        mol.getBondBetweenAtoms(monomer->getIdx(), bound_monomer->getIdx());
+    const auto& conf = mol.getConformer();
+    std::unordered_set<Direction> occupied_directions;
+
+    auto record_occupied_directions = [&](const RDKit::Bond* bond,
+                                          const bool is_secondary,
+                                          const std::string& linkage_property) {
+        std::string linkage;
+        if (!bond->getPropIfPresent(linkage_property, linkage) ||
```
KevKeating: I'm not sure that you need to confirm whether the property exists here.  `contains_two_monomer_linkages` already confirms that CUSTOM_BOND exists, and LINKAGE should always be present for a monomeric connection.

src/schrodinger/sketcher/rdkit/monomeric.cpp
```diff
@@ -220,24 +232,452 @@ does_connector_have_arrowheads(const RDKit::Bond* bond,
...
+        if (contains_two_monomer_linkages(bond)) {
+            record_occupied_directions(bond, true, CUSTOM_BOND);
+        }
+    }
+    return occupied_directions;
+}
+
+static bool monomers_are_in_same_chain(const RDKit::Atom* first,
+                                       const RDKit::Atom* second)
+{
```
KevKeating, line 358: You could write this as
```
return rdkit_extensions::get_polymer_id(first) == rdkit_extensions::get_polymer_id(second);
```
or as
```
try {
    return rdkit_extensions::get_polymer_id(first) == rdkit_extensions::get_polymer_id(second);
} catch (const std::runtime_error&) {
    return false;
}
```
if you wanted to protect against non-monomeric atoms.

src/schrodinger/sketcher/rdkit/monomeric.cpp
```diff
@@ -220,24 +232,452 @@ does_connector_have_arrowheads(const RDKit::Bond* bond,
...
+        const auto* end = bond->getEndAtom();
+        const QLineF bond_line(to_scene_xy(conf.getAtomPos(begin->getIdx())),
+                               to_scene_xy(conf.getAtomPos(end->getIdx())));
+        QPointF intersection;
+        if (candidate_line.intersects(bond_line, &intersection) ==
+                QLineF::BoundedIntersection &&
+            // Bonds ending at the remote monomer naturally meet the candidate
+            // at that monomer's center. Only interior intersections count.
+            !qFuzzyIsNull(QLineF(intersection, bound_coords).length())) {
+            ++crossing_count;
```
KevKeating: Is it worth trying to account for bonds containing secondary connections in this function?

src/schrodinger/sketcher/rdkit/monomeric.cpp
```diff
@@ -220,24 +232,452 @@ does_connector_have_arrowheads(const RDKit::Bond* bond,
...
+    const QGraphicsItem& monomer_item, const QPointF& bound_coords,
+    const RDKit::Atom* monomer, const RDKit::Atom* bound_monomer,
+    const std::vector<Direction>& directions)
+{
+    auto best_direction = directions.front();
+    auto fewest_crossings = count_candidate_bond_crossings(
+        monomer_item, bound_coords, monomer, bound_monomer, best_direction);
+    if (fewest_crossings == 0) {
+        return best_direction;
+    }
```
KevKeating: Are these lines needed?  Could you start the loop at `index = 0` instead and initialize `fewest_crossings` to `std::numeric_limits<size_t>::max()`?  (Or even just initialize it to `99999` or something like that, since something has gone horribly, horribly wrong if we have hundreds of thousands of line crossings.)

KevKeating (commented):
Nice!  I just had a few nitpicks.

ZontaNicola:
> > @ZontaNicola - Looking just at the gallery, I have one comment: 946 seems to have a gpa between the end of the connection and the cysteine monomer. I see that on 950 as well
> 
> It looks like the gap is due to the difference between the corner of the rectangle and the curve of the rounded rectangle. Presumably, adjusting diagonal coordinates by `AA_ROUNDING_RADIUS` would fix that.

yeah, also I'm wondering wether the diamond should be rotated when it is at the corners... I guess it's a question for Laura?

ZontaNicola:
Screenshot 2026-09-10 at 17 36 45
rotated diamond... I think I like this!

src/schrodinger/sketcher/molviewer/monomer_attachment_point_labels.cpp
ZontaNicola: Yes. I changed this to derive the normal from the connection vector. The reverse endpoint call reverses that vector, and therefore the normal, so the two labels stay on opposite sides even when both arrowheads use the same monomer side. Included in c718eb82.

src/schrodinger/sketcher/rdkit/monomeric.cpp
ZontaNicola: Agreed. I changed this helper to return a single Direction and updated its callers in c718eb82.

src/schrodinger/sketcher/rdkit/monomeric.cpp
ZontaNicola: Done in c718eb82. I moved direction_to_qt_vector into molviewer/coord_utils and now reuse it here, scaling diagonal vectors by COSINE_45_DEGREES.

src/schrodinger/sketcher/rdkit/monomeric.cpp
ZontaNicola: Yes, it meant the half facing the bound monomer rather than the back half. I reworded the comment explicitly in c718eb82.

src/schrodinger/sketcher/rdkit/monomeric.cpp
ZontaNicola: Agreed. I removed the redundant property-existence check and the now-unused linkage-property parameter in c718eb82.

src/schrodinger/sketcher/rdkit/monomeric.cpp
ZontaNicola: Done in c718eb82 using get_polymer_id, with the runtime_error catch so this remains safe if a non-monomeric atom reaches the helper.

src/schrodinger/sketcher/rdkit/monomeric.cpp
ZontaNicola: Yes. A bond containing a distinct secondary linkage is rendered as two connections now contributes a crossing penalty of two rather than one. Included in c718eb82.

src/schrodinger/sketcher/rdkit/monomeric.cpp
ZontaNicola: Agreed. The loop now starts with the first direction and fewest_crossings is initialized to std::numeric_limits<size_t>::max(). The early return for zero crossings is preserved. Included in c718eb82.

ZontaNicola:
Review feedback is addressed in c718eb82 and e7c4ad37. For the gallery 946/950 corner-gap feedback, the final version rotates diamonds by 45 degrees only at corners; side diamonds remain unchanged. I also tried offsetting corner coordinates by AA_ROUNDING_RADIUS, but gallery inspection showed that placed the diamonds too close, so that inset is not part of the final code. Local verification passed: Sketcher build, 10 monomeric tests, 4 coordinate-utility tests, and 17 monomeric-connection integration tests.

ZontaNicola (commented):
I let codex reply but cross checked everything and it looks good to me

KevKeating (approved):

ethan-schrodinger:
>  rotated diamond... I think I like this!

I think the rotation is a good idea when the diamond intersects with a corner, but in this case it still isn't quite intersecting with the monomer. Maybe it's not such a big deal, but it stands out a little to me.

src/schrodinger/sketcher/molviewer/monomer_connector_item.cpp
```diff
@@ -139,19 +155,29 @@ void MonomerConnectorItem::updateCachedData()
...
+    QPointF end_offset;
     if (end_has_arrowhead) {
-        end_pos.ry() -= get_monomer_arrowhead_offset(m_end_item, start_qcoords);
-        add_diamond_arrowhead_to_path(m_arrowhead_path, end_pos,
-                                      MONOMER_CONNECTOR_ARROWHEAD_RADIUS);
+        end_offset = get_monomer_arrowhead_offset(
+            m_end_item, start_qcoords, m_end_item.getAtom(),
+            m_start_item.getAtom(), m_is_secondary_connection);
     }
+    auto end_pos = end_qcoords - start_qcoords + end_offset;
```
ethan-schrodinger, line 169: If `end_has_arrowhead = False`, it looks like `end_offset` is uninitialized. Would it be safer to initialize?

ethan-schrodinger (approved):

src/schrodinger/sketcher/molviewer/monomer_connector_item.cpp
ZontaNicola: yeah

ZontaNicola:
I went for the rotated diamond which seems to please more people, but we can always revisit if we find a better option

PR #413, opened 2026-09-02
SKETCH-2643: Add a Playwright test bridge for the WASM sketcher

cdvonbargen:
- Linked Case: SKETCH-2643

### Description

I took Sean's PR, whined at claude a bunch, fed it Juzer comments, and ended up with this -- a small surface area of emscripten exposed functions that are able to collect coordinates of parts of the sketcher, or activate parts of the sketcher explicitly. I believe this should cover all of Sean's needs for atom/bond/menu/etc interaction. Below is claude's writeup of the changes:

Qt/WASM paints the whole app into one canvas, so Playwright has no DOM to target. Today's e2e tests work around that with two ad-hoc functions in `main.cpp` that only reach visible toolbar buttons; porting the rest of the Squish suite needs atoms, bonds, context menu rows, and the tools that only exist inside press-and-hold popups.

Instead of a binding per object kind, the bridge does the one thing Playwright can't do itself: resolve a selector (`widget:fit_btn`, `atom:2`, `bond:0`, `menu:Flip`) to page coordinates, along with enough state to tell whether Qt would accept a click. Tests take it from there with real mouse and keyboard events, so Qt's hit-testing stays in the loop and new selector kinds cost no new bindings. Two cases need more than geometry: a `QToolButton` menu runs a nested event loop that suspends the WASM stack under Asyncify, so its rows are activated rather than clicked, and a tool hidden in a popup needs its owning button pressed and held first — the bridge derives that owner from the widget tree rather than making tests carry a hand-written map of all seventeen.

This PR is infrastructure only. The ported Squish suites and their snapshots are held back for a follow-up branch so this one stays reviewable, which does mean most of the helper layer lands ahead of its callers — the existing e2e tests exercise the widget selector and nothing else. Worth flagging for review: like the two functions it replaces, the bridge is compiled unconditionally, so these bindings ship in the released WASM as they do today. They are underscore-prefixed, no production code path calls them, and the whole file compiles away outside Emscripten, but gating it behind a build flag would be a reasonable follow-up if we'd rather it not ship at all.

### Testing Done

CI. Locally, the ported bond context menu suite — held back for the follow-up branch — ran green against this bridge, exercising the `menu:`, `atom:`, and `bond:` selectors along with menu row activation. That run predates the review-comment commits that followed it, whose changes are covered by compilation only.

🤖 Generated with [Claude Code](https://claude.com/claude-code)

### DIFF
--- a/src/app/playwright_test_bridge.cpp
+++ b/src/app/playwright_test_bridge.cpp
@@ -48,6 +48,7 @@
 #include <QGraphicsView>
 #include <QJsonDocument>
 #include <QJsonObject>
+#include <QList>
 #include <QMenu>
 #include <QPoint>
 #include <QRect>
@@ -142,6 +143,10 @@ QWidget* find_visible_widget(SketcherWidget& sketcher, const QString& name)
  * held, so a test that wants such a tool has to know which button to hold.
  * Deriving that from the widget tree keeps tests from carrying a hand-written
  * map of every tool to its owning button.
+ *
+ * Should the same objectName appear in more than one popup, this returns the
+ * first owner found; name the widget unambiguously rather than relying on
+ * which one that is.
  */
 std::string popup_owner(SketcherWidget& sketcher, const std::string& name)
 {
@@ -223,7 +228,7 @@ QGraphicsItem* find_visible_item(const QGraphicsView& view, const bool is_atom,
  * "widget" only matches a visible widget, since that is what a test can click.
  * "state" matches whether or not the widget is showing and reports "visible",
  * which is how a test asserts that a shortcut selected a tool that lives in a
- * closed popup. A button also reports "checked" and "text".
+ * closed popup. A button also reports "checked", "text", and "toolTip".
  */
 std::string widget_rect(SketcherWidget& sketcher, const std::string& name,
                         const bool visible_only)
@@ -280,6 +285,29 @@ std::string item_rect(SketcherWidget& sketcher, const bool is_atom,
 }
 
 /**
+ * Find the action matching name_or_text, or nullptr if there is none.
+ *
+ * Most menu actions are created without an objectName, so tests usually name a
+ * row by the text the user sees. An objectName match still wins over a text
+ * match anywhere in the list, so that a test naming a specific action can't be
+ * captured by an unrelated one whose label happens to collide.
+ */
+QAction* find_action(const QList<QAction*>& actions, const QString& name)
+{
+    const QString wanted = without_mnemonic(name);
+    QAction* by_text = nullptr;
+    for (auto* action : actions) {
+        if (action->objectName() == name) {
+            return action;
+        }
+        if (by_text == nullptr && without_mnemonic(action->text()) == wanted) {
+            by_text = action;
+        }
+    }
+    return by_text;
+}
+
+/**
  * Resolve a row of one open QMenu, or "" if this menu has no matching row.
  *
  * Reading action geometry is only safe once the popup is actually visible; Qt
@@ -293,20 +321,16 @@ std::string action_rect_in_menu(const QMenu& menu,
     if (!menu.isVisible()) {
         return {};
     }
-    const QString wanted = without_mnemonic(name);
-    for (auto* action : menu.actions()) {
-        if (action->objectName() != name &&
-            without_mnemonic(action->text()) != wanted) {
-            continue;
-        }
-        const QRect rect = menu.actionGeometry(action);
-        if (rect.isEmpty()) {
-            continue;
-        }
-        return rect_to_json(map_to_sketcher(menu, sketcher, rect.topLeft()),
-                            rect.size(), action->isEnabled());
+    auto* action = find_action(menu.actions(), name);
+    if (action == nullptr) {
+        return {};
     }
-    return {};
+    const QRect rect = menu.actionGeometry(action);
+    if (rect.isEmpty()) {
+        return {};
+    }
+    return rect_to_json(map_to_sketcher(menu, sketcher, rect.topLeft()),
+                        rect.size(), action->isEnabled());
 }
 
 /**
@@ -339,26 +363,19 @@ std::string menu_rect(SketcherWidget& sketcher, const std::string& value)
 
 /**
  * Trigger the QAction matching name_or_text, or return false if there is none.
- *
- * Actions are matched on objectName first and then on display text, since most
- * menu actions are created without an objectName.
  */
 bool try_activate_action(SketcherWidget& sketcher, const QString& name)
 {
-    const QString wanted = without_mnemonic(name);
-    for (auto* action : sketcher.findChildren<QAction*>()) {
-        if (action->objectName() != name &&
-            without_mnemonic(action->text()) != wanted) {
-            continue;
-        }
-        if (!action->isEnabled()) {
-            throw std::runtime_error("playwright test bridge: action '" +
-                                     name.toStdString() + "' is disabled");
-        }
-        action->trigger();
-        return true;
+    auto* action = find_action(sketcher.findChildren<QAction*>(), name);
+    if (action == nullptr) {
+        return false;
     }
-    return false;
+    if (!action->isEnabled()) {
+        throw std::runtime_error("playwright test bridge: action '" +
+                                 name.toStdString() + "' is disabled");
+    }
+    action->trigger();
+    return true;
 }
 
 } // namespace
@@ -405,10 +422,11 @@ bool try_activate_action(SketcherWidget& sketcher, const QString& name)
  * order the structure was built in, not the canonical order a format like
  * SMILES may renumber to on export.
  *
- * Returns "{}" when nothing visible matches, since there is no coordinate a
- * test could click. "enabled" reports whether Qt would accept a click there;
- * Playwright applies that check automatically to real DOM elements but has no
- * way to do so for something painted into a canvas.
+ * Returns "{}" when nothing matches: for every selector but "state:" that means
+ * nothing visible matches, since there is no coordinate a test could click.
+ * "enabled" reports whether Qt would accept a click there; Playwright applies
+ * that check automatically to real DOM elements but has no way to do so for
+ * something painted into a canvas.
  *
  * Throws std::runtime_error for a malformed selector or an unrecognized kind,
  * which indicates the test needs to be updated.
--- a/src/app/playwright_test_bridge.cpp
+++ b/src/app/playwright_test_bridge.cpp
@@ -1,0 +1,491 @@
+/* -------------------------------------------------------------------------
+ * Playwright-only test bridge for the WebAssembly Sketcher application.
+ *
+ * Qt/WASM paints the entire application into a single canvas element, so
+ * Playwright cannot see individual widgets or scene items in the DOM. The
+ * bridge exists to close exactly that gap and no more: sketcher_get_rect()
+ * resolves a selector to canvas coordinates, and the test then drives the
+ * application with real Playwright mouse and keyboard events. Anything a test
+ * can do with coordinates plus Playwright belongs in the JavaScript helpers,
+ * not here.
+ *
+ * A Qt::Popup widget is painted into its own canvas element, but that element
+ * is positioned over the page, so a rect mapped through global coordinates is
+ * still a clickable page coordinate. Such a widget is reached with a real mouse
+ * event like anything else.
+ *
+ * A context menu behaves the same way: the sketcher shows one with
+ * QMenu::show() rather than exec(), so it does not run a nested event loop and
+ * its rows are located with a "menu:" selector and clicked for real.
+ *
+ * A QToolButton's menu is the one exception, and sketcher_activate_action()
+ * exists for it alone: Qt does run a nested event loop while such a menu is
+ * open, which under Asyncify suspends the WebAssembly stack and stops any call
+ * into the module from completing until the menu closes. See that function.
+ *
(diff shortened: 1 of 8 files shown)


Sean-Seekins:
Now I can take this PR and feed it directly into my own forked project to see if I can implement this logic 👍

Sean-Seekins:
Update on work so far (from agent):

Your generic geometry-bridge suggestion appears to be the right next architectural step. My first PR established that the Squish workflows can be reproduced in Playwright with real browser interaction; the generic bridge is a cleaner way to make that sustainable by replacing many test-specific object bindings with a small reusable geometry/state interface.
Early validation supports the approach. We are still completing regression validation, but nothing so far suggests the generic bridge is incompatible with the converted suite. The remaining work is compatibility hardening around a few Qt/WASM popup and dialog surfaces, not a reason to return to a growing per-object bridge model.
Separately, I kept all test-related C++ behind an opt-in WASM test-build flag because QA GUI automation normally lives outside the application repository. That preserves a clean production build while letting the Playwright work evolve here; I’m happy to adjust that boundary if the team prefers a different ownership model.

Sean-Seekins:
Another follow up.  This is essentially done apart from a few small things to make this work:

I staged the converted Squish-parity suite on top of this bridge. The shared geometry/state approach works well for the pointer-driven wrappers, and the full current regression suite passes (51/51) on the integration branch.

To cover the remaining cases without returning to bespoke widget bridges, the follow-up branch includes generic bridge hardening for active dialogs/popups and standard control state, plus two additional reusable capabilities:

- geometry lookup for a visible `QComboBox` popup item by text, so Playwright can select export formats with a real mouse click;
- passive capture of Qt/WASM file-export bytes after the test performs the normal Download click, since that path does not produce a normal browser download event.

I’ve separated the work into two review branches:

- [generic bridge follow-up](https://github.com/Sean-Seekins/sketcher/tree/playwright-bridge-followup)
- [stacked parity tests/wrappers](https://github.com/Sean-Seekins/sketcher/tree/playwright-parity-tests)

My plan is to open the generic bridge extension as a narrow follow-up PR after this lands, then make the downstream Sean PR tests/wrappers/fixtures/references only. Does that separation fit the intended direction?

src/app/sketcher_instance.h
```diff
@@ -0,0 +1,20 @@
+/* -------------------------------------------------------------------------
+ * Access to the application's single SketcherWidget.
+ *
+ * Defined in main.cpp; declared here so that other translation units in the
+ * app can reach the instance without duplicating the declaration.
+ *
+ * Copyright Schrodinger LLC, All Rights Reserved.
```
ethan-schrodinger: FWIW, Schrodinger is no longer an LLC, and separately, this type of copyright notice is no longer necessary.

test/wasm/e2e/e2e_helpers.js
```diff
@@ -86,26 +196,388 @@ export async function isSketcherEmpty(page) {
...
+  }
+  try {
+    await page.mouse.move(x, y, { steps: 4 });
+    await page.mouse.down({ button });
+    await page.mouse.up({ button });
+  } finally {
+    for (const modifier of [...modifiers].reverse()) {
+      await page.keyboard.up(modifier);
+    }
+  }
```
juzerzarif, line 308: Should there be a `hideMouseMarker` here in this `finally` block

test/wasm/e2e/e2e_helpers.js
```diff
@@ -86,26 +196,388 @@ export async function isSketcherEmpty(page) {
...
+ * @param {string} name - Qt objectName of the widget
+ * @param {object} [options] - see mouseClick
+ */
+export async function clickWidget(page, name, options) {
+  await click(page, `widget:${name}`, options);
+}
+
+// ToolButtonWithPopup::m_popup_delay is 250 ms, and releasing before it elapses
+// cancels the popup, so hold well past it. Raise this if that default changes.
+const POPUP_HOLD_MS = 600;
```
juzerzarif, line 411: Probably worth putting an equivalent comment on `ToolButtonWithPopup::m_popup_delay` to remind people to change these two values together.

test/wasm/e2e/e2e_helpers.js
```diff
@@ -86,26 +196,388 @@ export async function isSketcherEmpty(page) {
...
+}
+
+/**
+ * Move the pointer over whatever a selector resolves to.
+ *
+ * @param {import('@playwright/test').Page} page
+ * @param {string} selector - see getRect
+ */
+async function hover(page, selector) {
+  const rect = await waitForClickable(page, selector);
```
juzerzarif, line 464: Don't only buttons return the `enabled` state? You might wanna hover on something other than a button right?

juzerzarif (approved):
A few minor nits but the API shape for the bridge LGTM 👍

cdvonbargen:
Addressed Juzer's comments, and incorporated reporting back text from widget_rect() (per Sean's branch)

I think other things on his branch can be simplified, but let's get this in first and we can move the conversation over there.

src/app/playwright_test_bridge.cpp
```diff
@@ -0,0 +1,531 @@
...
+ * is positioned over the page, so a rect mapped through global coordinates is
+ * still a clickable page coordinate. Such a widget is reached with a real mouse
+ * event like anything else.
+ *
+ * A context menu behaves the same way: the sketcher shows one with
+ * QMenu::show() rather than exec(), so it does not run a nested event loop and
+ * its rows are located with a "menu:" selector and clicked for real.
+ *
+ * A QToolButton's menu is the one exception, and sketcher_activate_action()
+ * exists for it alone: Qt does run a nested event loop while such a menu is
```
KevKeating: Nit: changing "does run" to "runs" seems like it would be a little clearer.

src/app/playwright_test_bridge.cpp
```diff
@@ -0,0 +1,531 @@
...
+ *
+ * The casts are dynamic_casts to the abstract base classes rather than
+ * qgraphicsitem_casts to AtomItem/BondItem, because qgraphicsitem_cast matches
+ * type() exactly and so would skip monomers and monomer connectors: those are
+ * siblings of AtomItem and BondItem, not subclasses.
+ */
+QGraphicsItem* find_visible_item(const QGraphicsView& view, const bool is_atom,
+                                 const int index)
+{
+    for (auto* item : view.scene()->items()) {
```
KevKeating: You could probably simplify this implementation a bit by calling `getIntertiveItems` with `InteractiveItemFlag::ATOM_OR_MONOMER` and/or `InteractiveItemFlag::BOND_OR_CONNECTOR` instead of items.  That would also avoid the risk of accidentally grabbing a graphics item from a predictive hint that's temporarily shown on hover.

src/app/playwright_test_bridge.cpp
```diff
@@ -0,0 +1,531 @@
...
+/**
+ * Resolve a "widget:<objectName>" or "state:<objectName>" selector, or "{}" if
+ * nothing matches.
+ *
+ * "widget" only matches a visible widget, since that is what a test can click.
+ * "state" matches whether or not the widget is showing and reports "visible",
+ * which is how a test asserts that a shortcut selected a tool that lives in a
+ * closed popup. A widget whose text the user can read reports it as "text", and
+ * a button also reports "checked" and "toolTip".
+ */
```
KevKeating, line 264: I found this docstring to be really confusing.  `"widget:<objectName>" or "state:<objectName>" selector` refer to strings that aren't actually sent into that function, so I wasn't sure what they were referring to here.  The docstring also says that `"widget" only matches a visible widget`, but doesn't explain how that fits in with the `visible_only` parameter.  I think the docstring should be rewritten to explain how the function responds to _its_ input, not to Playwright input.  Also, the docstring should at least mention that the function returns JSON.

src/app/playwright_test_bridge.cpp
```diff
@@ -0,0 +1,531 @@
...
+        if (widget == nullptr) {
+            return "{}";
+        }
+    }
+    auto result = rect_json(map_to_sketcher(*widget, sketcher, QPoint(0, 0)),
+                            widget->size(), widget->isEnabled());
+    if (!visible_only) {
+        result["visible"] = widget->isVisible();
+    }
+    // A caption is chrome, so it is normalized the way the user reads it; a
```
KevKeating: It took me a while to figure out what "caption" meant here, since that's not how Qt refers to anything.  My first guess was that it referred to tool tips or something like that.

src/app/playwright_test_bridge.cpp
```diff
@@ -0,0 +1,531 @@
...
+    auto* widget = find_visible_widget(sketcher, object_name);
+    if (widget == nullptr) {
+        if (visible_only) {
+            return "{}";
+        }
+        widget = sketcher.findChild<QWidget*>(object_name);
+        if (widget == nullptr) {
+            return "{}";
+        }
+    }
```
KevKeating, line 275: Is this search intentionally preferring visible widgets over hidden ones of the same name?  I think it would be more readable as something like
```
    QWidget* widget
    if (visible_only) {
        widget = find_visible_widget(sketcher, object_name);
    } else {
        widget = sketcher.findChild<QWidget*>(object_name);
    }
    if (widget == nullptr) {
        return "{}";
    }
```

src/app/playwright_test_bridge.cpp
```diff
@@ -0,0 +1,531 @@
...
+    return to_json(result);
+}
+
+/**
+ * Resolve an "atom:<index>" or "bond:<index>" selector, or "{}" if no visible
+ * item matches. Scene items are QGraphicsItems rather than QWidgets, so they
+ * can't be found by objectName; their bounding rect is mapped through the View
+ * transform instead.
+ */
```
KevKeating, line 315: This docstring is really confusing as well.  It would be useful to have `param` blocks to explain that `is_atom` distinguishes between atoms and bonds, and that `value` is supposed to be a string containing an integer.

Is it worth creating an `AtomOrBond` enum with `ATOM` and `BOND` values to replace the `is_atom` bool?  I find `is_atom` to be confusing since it's not immediately obvious that it's distinguishing atoms from bonds, as opposed to atoms from monomers, or atoms from R-groups, etc.

src/app/playwright_test_bridge.cpp
```diff
@@ -0,0 +1,531 @@
...
+    }
+    const QRect viewport_rect =
+        view.mapFromScene(item->sceneBoundingRect()).boundingRect();
+    return rect_to_json(
+        map_to_sketcher(*view.viewport(), sketcher, viewport_rect.topLeft()),
+        viewport_rect.size(), item->isEnabled());
+}
+
+/**
+ * Find the action matching name_or_text, or nullptr if there is none.
```
KevKeating: I think this would be a little clearer if you added "in the list of actions" after "name_or_text".  Also, the parameter is called "name", not "name_or_text", although "name_or_text" seems like a better name.

src/app/playwright_test_bridge.cpp
```diff
@@ -0,0 +1,531 @@
...
+    return by_text;
+}
+
+/**
+ * Resolve a row of one open QMenu, or "" if this menu has no matching row.
+ *
+ * Reading action geometry is only safe once the popup is actually visible; Qt
+ * has not laid the rows out before that, and on WASM touching a menu's
+ * internals mid-show aborts the runtime.
+ */
```
KevKeating, line 376: This docstring should explicitly point out that this function can return an empty string, since most of the other functions return "{}" if there's no data to return, and the distinction between `{}` and `"{}"` isn't exceedingly obvious at a glance.

src/app/playwright_test_bridge.cpp
```diff
@@ -0,0 +1,531 @@
...
+}
+
+/**
+ * Trigger the QAction matching name_or_text, or return false if there is none.
+ */
+bool try_activate_action(SketcherWidget& sketcher, const QString& name)
+{
+    auto* action = find_action(sketcher.findChildren<QAction*>(), name);
+    if (action == nullptr) {
+        return false;
```
KevKeating: Why not just have this function throw an exception here instead of having `sketcher_activate_action` do it instead?

src/app/playwright_test_bridge.cpp
```diff
@@ -0,0 +1,531 @@
...
+        "' (expected 'widget', 'state', 'atom', 'bond', or 'menu')");
+}
+
+/**
+ * Trigger a menu action by objectName or visible text, without opening the menu
+ * it belongs to. Mnemonic ampersands are ignored when comparing text.
+ *
+ * This exists because a menu row cannot be clicked the way every other control
+ * can. The menu belongs to a QToolButton, and Qt runs a nested event loop for
+ * as long as that menu is open. Qt/WASM is built with Asyncify, so that nested
```
KevKeating, line 518: There's a chance that this restriction will be removed with Qt 6.10 or later, assuming we can ever get our JavaScript functions working there.  (There's nothing that needs doing for this comment; just wanted to mention it.)

KevKeating (commented):
Overall, the C++ code looks good to me.  I had a handful of comments, but most of them are just about docstrings.  I glanced through the JavaScript code, but I don't know enough JavaScript to be of much help there.

src/app/playwright_test_bridge.cpp
cdvonbargen: yeah, claude pushed back on this because it does want visible-first preference; the visible one is the only clickable match.

KevKeating (approved):

ethan-schrodinger (approved):

PR #419, opened 2026-09-04
WIP: Port Sketcher parity tests to Playwright

Sean-Seekins:
## Status

Draft review of the Squish-to-Playwright Sketcher e2e port. This is intentionally not ready to merge yet.

## Scope

- Converted browser-facing e2e suites and source-faithful Playwright wrappers
- Versioned runtime fixtures and per-test archived Playwright visual baselines
- Focused wrapper/geometry regression contracts
- A provisional generic Qt/WASM geometry bridge used to validate the port

## Validation

- Integrated locally with current `main` at `9a69fd1d`
- Rebuilt the local WASM artifact
- Full Playwright e2e suite: **77 passed**

## Follow-up with #413

PR #413 provides the intended shared generic Playwright geometry bridge. After it lands, this branch will be integrated against it and redundant bridge code removed. The remaining PR should then consist primarily of the converted tests, parity wrappers, fixtures, and any genuinely necessary test-specific support.

Reviewers can focus on the test-port structure and source parity now; bridge reconciliation will be a separate, verified follow-up.


Sean-Seekins:
One architecture note: this draft uses an independently validated implementation of the generic-geometry approach from #413, but it has not yet incorporated #413’s commits directly.

The intended end state is the same small shared interface: generic selector-based geometry/state lookup, popup-owner lookup, and the narrow QAction activation exception required by Qt/WASM’s nested tool-menu behavior. Normal test actions use real Playwright mouse and keyboard input.

This draft still retains the previous bridge exports because the Squish-to-Playwright port was validated incrementally and they provide a rollback path while we establish parity. A few source-parity helpers also still use legacy diagnostics such as rendered atom/bond geometry and visible-widget discovery.

After #413 lands, I plan to re-integrate this work against that implementation, replace the overlapping bridge code, convert removable direct helpers to ordinary Playwright interactions, and identify whether the remaining read-only diagnostic cases need a small shared extension or a test rewrite.

The current draft is therefore mainly for review of the converted tests, source-faithful wrappers, fixtures, and reference organization. It has passed the full local suite against current main (77 tests).

PR #422, opened 2026-09-09
Fix buildinger sketcher build and test failures on macOS

cdvonbargen:
- Linked Case: N/A

### Description

`buildinger sketcher -i --with-tests` fails on macOS because premake hardcoded the package factory environment name as `schrodinger`, while macOS now defaults to `schrodinger-gfortran`; this uses `SCHRODINGER_LIB` instead, which build_env already exports for whichever environment was selected. With that fixed, every C++ test aborted at load because the prebuilt `libRDKitChemDraw` needs libexpat from the environment's top level lib directory, which CMake never adds to the rpath since nothing links against it, so that directory is now added to `CMAKE_BUILD_RPATH`.

### Testing Done

Confirmed `buildinger sketcher -i --with-tests` now builds and passes all 82 tests, the same 82 that previously aborted with the dyld error.

### DIFF
--- a/premake
+++ b/premake
@@ -23,6 +23,9 @@ BUILD_DIR=$SCHRODINGER/sketcher
 # SCHRODINGER_LIB is exported by build_env and points at the package factory
 # environment selected for this build (e.g. schrodinger, schrodinger-gfortran).
 PF_DIR=$SCHRODINGER_LIB
+if [[ "$OSTYPE" == "msys" ]]; then
+    PF_DIR=$(cygpath -m $PF_DIR)
+fi
 ZLIB_DIR=$(ls -d $PF_DIR/zlib-*)
 EIGEN_DIR=$(ls -d $PF_DIR/share/eigen3)
 COORDGEN_DIR=$(ls -d $PF_DIR/coordgenlibs-*/lib/cmake)
--- a/premake
+++ b/premake
@@ -20,7 +20,9 @@ set -euo pipefail
 # Define directory paths
 SRC_DIR=$SCHRODINGER_SRC/sketcher
 BUILD_DIR=$SCHRODINGER/sketcher
-PF_DIR=$SCHRODINGER/schrodinger_buildenv_packages/.pixi/envs/schrodinger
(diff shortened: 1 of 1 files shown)


premake
```diff
@@ -20,7 +20,9 @@ set -euo pipefail
 # Define directory paths
 SRC_DIR=$SCHRODINGER_SRC/sketcher
 BUILD_DIR=$SCHRODINGER/sketcher
-PF_DIR=$SCHRODINGER/schrodinger_buildenv_packages/.pixi/envs/schrodinger
+# SCHRODINGER_LIB is exported by build_env and points at the package factory
+# environment selected for this build (e.g. schrodinger, schrodinger-gfortran).
+PF_DIR=$SCHRODINGER_LIB
```
KevKeating, line 25: On Windows, I need to change this line to
```suggestion
PF_DIR=$SCHRODINGER_LIB
if [[ "$OSTYPE" == "msys" ]]; then
    PF_DIR=$(cygpath -m $PF_DIR)
fi
```
in order for builds to succeed.  Otherwise, I get
```
CMake Error at CMakeLists.txt:36 (find_package):
  By not providing "Findmaeparser.cmake" in CMAKE_MODULE_PATH this project
  has asked CMake to find a package configuration file provided by
  "maeparser", but CMake did not find one.

  Could not find a package configuration file provided by "maeparser" with
  any of the following names:

    maeparserConfig.cmake
    maeparser-config.cmake

  Add the installation prefix of "maeparser" to CMAKE_PREFIX_PATH or set
  "maeparser_DIR" to a directory containing one of the above files.  If
  "maeparser" provides a separate development package or SDK, be sure it has
  been installed.


-- Configuring incomplete, errors occurred!
```
because $SCHRODINGER_LIB is a Unix-style path, while $SCHRODINGER is a Windows-style path.  (Even though my shell defines $SCHRODINGER as a Unix-style path.  Presumably, buildinger auto-converts $SCHRODINGER but not $SCHRODINGER_LIB before calling into premake?)

KevKeating (approved):

PR #425, opened 2026-09-10
SKETCH-2643: Resolve widgets on foreground Qt surfaces

Sean-Seekins:
## Summary

- Resolve visible widgets on the foreground Qt surface first: active popup,
  active modal, then other visible top-level widgets before the background
  Sketcher widget tree.
- Include a widget's Qt style sheet in generic widget-state results so tests can
  inspect user-visible validation styling.
- Preserve the press-and-hold mouse gesture while activating a tool inside a
  popup, matching the interaction a user performs.
- Add focused Playwright contracts for the Paste in Text dialog and a periodic
  table embedded in a context menu.

## Why

PR #413 provides the generic selector-to-geometry bridge needed by the
Playwright tests. Some dialogs and QWidgetAction contents are separate
top-level Qt surfaces rather than children of SketcherWidget. A same-named
background widget could therefore be resolved instead of the foreground
control, or the foreground control could be missed entirely.

This is a narrow follow-up to make those foreground interactions work through
the same generic bridge. It does not add specialized bridge functions or CMake
changes.

## Validation

- Focused bridge contracts: 2 passed
- Stacked validation with the converted suite: all 51 suite tests passed
- JavaScript syntax and git diff checks passed

This bridge follow-up is intended to land before the larger tests-only change,
so that the latter can remain test and test-configuration code.

### DIFF
--- a/test/wasm/e2e/playwright_bridge_top_level.test.js
+++ b/test/wasm/e2e/playwright_bridge_top_level.test.js
@@ -1,0 +1,38 @@
+import { expect, test } from '@playwright/test';
+import {
+  activateAction,
+  clickPopupTool,
+  getExportedSmiles,
+  loadStructure,
+  openContextMenu,
+  selectAll,
+  waitForSketcherReady,
+  widgetState,
+} from './e2e_helpers.js';
+
+const SOURCE = 'NC(N)=NC(=O)CC1=C(Cl)C=CC=C1Cl';
+
+test.describe('Playwright bridge top-level Qt surfaces', () => {
+  test('reports text and style from the foreground Paste in Text dialog', async ({ page }) => {
+    await waitForSketcherReady(page);
+    await loadStructure(page, SOURCE);
+    await activateAction(page, 'Paste in Text...');
+
+    const status = await widgetState(page, 'status_lbl');
+    expect(status.text).toBe('Specified structure will <b>replace</b> Sketcher content');
+    expect(status.styleSheet).toBe('color : #c87c00');
+  });
+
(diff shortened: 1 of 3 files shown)


test/wasm/e2e/playwright_bridge_top_level.test.js
```diff
@@ -0,0 +1,38 @@
+import { expect, test } from '@playwright/test';
```
cdvonbargen, line 1: would it make sense to incorporate this into a more specific test? 

I'm trying to figure out the value of a generic infra test here (since we have all the other infra exercised through specific tests) -- curious what @KevKeating thinks, one thought I had was check this in and then kill it once it's used by the "real" tests

Sean-Seekins: TLDR- Just an intermediate infra test for faster production.  Can probably be moved somewhere else. 

Details - I created some infra tests along the way to save a bit of time with developing the wrappers.  That just let me run a quick set of tests to validate changes instead of running the full set of GUI tests.  I agree that we could move this to a specific test.

In general I think we can reduce a lot of testing overlap with this system.  I want to get the squish tests fully converted, 1-1, first to give us a clean starting point that matches the squish setup.  After that we can do a lot of refactoring and consolidating.

KevKeating: >  I want to get the squish tests fully converted, 1-1, first to give us a clean starting point that matches the squish setup. After that we can do a lot of refactoring and consolidating.

That seems like a reasonable approach to me.

KevKeating (approved):

cdvonbargen (approved):

PR #426, opened 2026-09-11
SKETCH-2788: Adding "Load Monomer Database..." to import button menu

KevKeating:
- Linked Case: SKETCH-2788

### Description

This PR adds a new "Load Monomer Database..." entry to the menu of the top bar's import button.  As part of this, I added a function to update the side bar monomer buttons so that the pop ups reflect the new monomer database.  This new function is now triggered when updating the monomer database through the new menu item or through the monomer DB functions defined in `main.cpp`.

I tested the new menu item with the JSON file that Rachel attached to the Jira case.  Initially, the monomer database complained about the ID field, but everything worked correctly once I edited the file to remove the "ID:" lines.  However, the side bar pop ups don't currently do a _great_ job of displaying large numbers of analogs:
image

(You might need to click on that image to zoom in.  The larger white thing on the right-hand side is the Sketcher window.  The long narrow strip is the pop up for the glycine button.)  I've filed SKETCH-2858 to improve this.  Alternatively, we can add a bullet point to the release notes that the new monomeric Sketcher takes full advantage of ultra-widescreen displays and multi-monitor setups.

### Testing Done

Tested on Windows desktop app

### DIFF
--- a/include/schrodinger/sketcher/sketcher_widget.h
+++ b/include/schrodinger/sketcher/sketcher_widget.h
@@ -121,6 +121,11 @@ class SKETCHER_API SketcherWidget : public QWidget
     void clear();
 
     /**
+     * Refresh sidebar monomer buttons after changing the monomer database.
+     */
+    void updateMonomerButtons();
+
+    /**
      * @return true if the scene is empty
      */
     bool isEmpty() const;
--- a/src/app/main.cpp
+++ b/src/app/main.cpp
@@ -113,6 +113,7 @@ void sketcher_load_custom_monomers_from_sql(const std::string& sql)
 {
     auto& db = schrodinger::rdkit_extensions::MonomerDatabase::instance();
     db.loadMonomersFromSql(sql);
+    get_sketcher_instance().updateMonomerButtons();
 }
 
 #ifdef __EMSCRIPTEN__
@@ -142,6 +143,7 @@ emscripten::val sketcher_load_custom_monomers(const std::string& json)
 {
     auto& db = schrodinger::rdkit_extensions::MonomerDatabase::instance();
     auto result = db.loadMonomersFromJson(json);
+    get_sketcher_instance().updateMonomerButtons();
 
     return monomer_defs_insertion_result_to_js(result);
 }
@@ -150,6 +152,7 @@ emscripten::val sketcher_insert_custom_monomers(const std::string& json)
 {
     auto& db = schrodinger::rdkit_extensions::MonomerDatabase::instance();
     auto result = db.insertMonomersFromJson(json);
+    get_sketcher_instance().updateMonomerButtons();
 
     return monomer_defs_insertion_result_to_js(result);
 }
@@ -159,6 +162,7 @@ void sketcher_reset_custom_monomers()
 {
     auto& db = schrodinger::rdkit_extensions::MonomerDatabase::instance();
     db.resetMonomerDefinitions();
+    get_sketcher_instance().updateMonomerButtons();
 }
 
 void sketcher_changed()
--- a/src/schrodinger/sketcher/menu/sketcher_top_bar_menus.cpp
+++ b/src/schrodinger/sketcher/menu/sketcher_top_bar_menus.cpp
@@ -49,6 +49,8 @@ ImportMenu::ImportMenu(QWidget* parent) : QMenu(parent)
     m_paste_in_text_act = addAction("Paste in Text...");
     m_replace_content_act = addAction("Replace Current Content");
     m_replace_content_act->setCheckable(true);
+    addSeparator();
+    m_load_monomer_database_act = addAction("Load Monomer Database...");
 }
 
 ExportMenu::ExportMenu(QWidget* parent) : QMenu(parent)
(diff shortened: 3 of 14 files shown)


KevKeating:
I just tested this with a WASM build, and everything works correctly except for message box reporting a successful read.  I'm assuming we need our own `MessageBox` class in the style of `ErrorDialog`, i.e. it needs to inherit from `ModalDialog` so that it works correctly with WASM.  I'll implement that next week.  I think it's worth keeping the message box, since otherwise it looks like absolutely nothing happened after the user selected their file.

ZontaNicola (approved):
This looks good to me... the related case of avoiding the huge panel seems easy enough that maybe I'd suggest putting it together in one single PR?

ethan-schrodinger (approved):
>Alternatively, we can add a bullet point to the release notes that the new monomeric Sketcher takes full advantage of ultra-widescreen displays and multi-monitor setups.

You're always teaching me new things about the art of engineering.
>This looks good to me... the related case of avoiding the huge panel seems easy enough that maybe I'd suggest putting it together in one single PR?

If we have a trivial fix that makes this look even a little better, it might not be a bad idea. At the very least it would make certification easier if QA is able to see what these entries are.

include/schrodinger/sketcher/sketcher_widget.h
```diff
@@ -120,6 +120,11 @@ class SKETCHER_API SketcherWidget : public QWidget
      */
     void clear();
 
+    /**
+     * Refresh sidebar monomer buttons after changing the monomer database.
+     */
+    void updateMonomerButtons();
```
cdvonbargen, line 126: do we want this as part of the public API? is it crazy to have the sketcher_model (or mol_model) listen to whether the monomerDb has been changed?

KevKeating: As far as I know, Monomer DB doesn't announce when it changes.  It's not a QObject, so we can't just add a signal.  We could modify things so that all monomer DB changes have to go through the `SketcherWidget`, but that would presumably require adding `loadMonomersFromSql`, `loadMonomersFromJson`, `resetMonomerDefinitions`, etc. methods to `SketcherWidget`.  We could modify Monomer DB to accept a callback when the monomers change and/or add a Qt wrapper that handles signals, but neither of those would be trivial changes (although they wouldn't be _all_ that hard).

KevKeating:
I've now updated the PR with a number of changes:

- The dialog confirming that the monomer database was updated successfully now works correctly on WASM
- The monomer pop ups now use multiple rows of buttons instead of a single row so they'll fit on a single monitor
- The monomer database now emits a signal when the database is updated and the Sketcher uses that signal to update the monomer pop ups.  As a result, there's no more need for the `SketcherWidget::updateMonomerButtons` method, so I've removed it.  I didn't want to add a Qt dependency to rdkit_extensions just for this signal, so Codex suggested using Boost's signal and slot implementation.  It requires adding another Boost library (`signals2`) to `BOOST_LIBS_TO_INCLUDE` in `external/CMakeLists.txt`, but the library is header-only so nothing needs to be built or linked.  I double checked this on both a Windows desktop build and a WASM build, and it works correctly on both.  Chris, does this approach seem reasonable to you?
- The new tests no longer modify the real on-disk monomer database by using the existing LocalMonomerDbFixture.  This led to a crash during test_sketcher_widget teardown, since test_sketcher_widget intentionally leaks `TestWidgetFixture`, and this leaked widget would attempt to respond to LocalMonomerDbFixture reseting the database _after_ the QApplication had been destroyed.  Codex fixed this by modifying `TestWidgetFixture`, and by making `QApplicationRequiredFixture` responsible for invoking `LocalMonomerDbFixture` for the GUI tests.  (`LocalMonomerDbFixture` is still applied to the `rdkit_extensions` tests without `QApplicationRequiredFixture`.)

include/schrodinger/sketcher/sketcher_widget.h
KevKeating: Update: I added a Boost signal that the Sketcher uses to update.  This avoids adding a Qt dependency to rdkit_extensions.  See my comment below for more details.

src/schrodinger/sketcher/widget/monomer_symbol_popup_utils.cpp
```diff
@@ -13,13 +13,39 @@ namespace schrodinger
 namespace sketcher
 {
 
+/**
+ * Determine the number of columns we should use for a monomer popup containing
+ * the specified number of monomers. This function tries to keep the popup
+ * roughly square-ish while using the same number of columns for popups with
+ * similar numbers of monomers.
+ */
+static size_t get_num_columns(const size_t num_monomers)
```
ethan-schrodinger, line 22: FWIW, I asked Laura about this and she seemed to say that anything we get from the monomer database should stay in the monomer database dialog and not be used to populate the popups of individual buttons (see [this comment](https://schrodinger.atlassian.net/browse/SKETCH-2858?focusedCommentId=4289736)). Given that, I don't think we'll need to plan for very large popups.

ethan-schrodinger (approved):

cdvonbargen (approved):
yessssss

rachelnwalker (approved):

PR #428, opened 2026-09-14
Expose sketcher APIs to cross-origin embeds

d-b-w:
- Linked Case: N/A

### Description

Applications that host the sketcher in an iframe currently cannot use its molecule APIs when the sketcher and host are served from different origins. This change adds a request/response `postMessage` protocol so a parent application can populate the sketcher, retrieve the current molecule, and render an image without direct access to the iframe's JavaScript context.

The protocol exposes `setMolecule`, `getMolecule`, and `getImage` with correlated request IDs and structured errors. The page waits for WASM initialization, accepts requests only from its immediate parent, and replies only to that parent's verified origin. Same-origin applications can call equivalent promise-based methods directly.

New WASM builds export Emscripten's stack and exception reference-count helpers so malformed molecule requests restore stack state and release native exception storage. The shell and compiled WASM assets are expected to come from the same distribution.

Detailed integration instructions and a complete example live in `wasm/README.md`, with a short link from the top-level README.

### Testing Done

- Ran 8 focused API tests through Playwright covering readiness, molecule import/export, image export, message validation, correlated responses, errors, and Emscripten exception cleanup.
- Ran 5 Playwright tests with the parent and iframe on different origins against the released 2026-3 WASM assets before removing mixed-version runtime compatibility. Verified direct iframe access is blocked while `postMessage` supports SMILES and MOL V3000 round trips, PNG and SVG export and browser decoding, clearing, invalid formats, and recovery after four malformed SMILES inputs.
- Ran Prettier checks and `git diff --check`.

Co-authored-by: Codex noreply@openai.com


README.md
```diff
@@ -21,6 +21,111 @@ This project is released by Schrödinger, Inc. and is available under an open-so
 
 **[Training and Video Walkthrough](https://www.schrodinger.com/sites/default/files/s3/public/2D-Sketcher/2023-2/Content/Resources/Videos/2D_Sketcher.mp4)** -- Learn how to use Sketcher with this guided video.
 
+## Embedding the HTML page
```
cdvonbargen, line 24: If you want this README, can you bury it under wasm/ subdir? Doesnt seem like this verbosity fits with the toplevel README

test/wasm/package.json
```diff
@@ -5,7 +5,8 @@
     "type": "module",
     "scripts": {
         "postinstall": "playwright install --with-deps",
-        "test": "playwright test"
+        "test": "node --test html_api.node.js && playwright test",
+        "test:html-api": "node --test html_api.node.js"
```
cdvonbargen, line 9: why is this needed and not just run through `playwright test` itself?

README.md
```diff
@@ -21,6 +21,111 @@ This project is released by Schrödinger, Inc. and is available under an open-so
...
+URL prefix. Use `data:image/png;base64,` for PNG or
+`data:image/svg+xml;base64,` for SVG when displaying it in an image element.
+
+Only the immediate embedding parent may send requests. Replies target the
+sender's exact origin. Unrelated messages and requests without a string ID are
+ignored. Serve both applications over HTTP(S); opaque parent origins such as
+sandboxed pages without `allow-same-origin` are not supported. Hosting headers
+must permit framing by your application.
+
+Same-origin hosts can also call the promise-based `window.getMolecule(format)`, `window.getImage(format)`, and `window.setMolecule(text)` methods directly. The existing `window.Module`
```
KevKeating, line 122: This line looks like it needs to be re-wrapped.

README.md
```diff
@@ -21,6 +21,111 @@ This project is released by Schrödinger, Inc. and is available under an open-so
...
+API remains available for advanced use.
+
+Older WASM distributions without the exception cleanup exports remain compatible,
```
KevKeating, line 125: "compatible" with what?  Is this saying you could use an older WASM with a newer wasm_shell.html?  Is that something that's worth documenting?  I can't think of many scenarios where there'd be much benefit to doing that.  Is it worth supporting?

KevKeating (commented):
> restricts callers to its immediate parent

I'm assuming this is for security reasons?

Also, out of curiosity, do you know if switching from asyncify to JSPI would make any of this easier and/or break anything here (other than the obvious caveat of not working on older browsers that don't support JSPI)?

juzerzarif:
Just curious what use case is this for??

README.md
d-b-w: Moved the detailed guide to `wasm/README.md` and left only a short link in the top-level README.

test/wasm/package.json
d-b-w: Agreed. The API tests now import Playwright’s test function and run as part of the normal `playwright test` command; the separate Node test script is gone.

README.md
d-b-w: Fixed as part of moving and formatting the guide under `wasm/`.

README.md
d-b-w: Agreed that mixing a newer shell with older WASM assets is not a useful supported configuration. I removed the compatibility path and this documentation; the shell now requires the runtime exports built with the same distribution.

d-b-w:
The use case is a web application that embeds the hosted sketcher from another origin and needs to initialize it from an existing molecule, retrieve the edited structure, or request a rendered image. Same-origin callers could already reach `window.Module`; this protocol provides those operations without requiring both applications to share an origin. I updated the PR description and `wasm/README.md` to state that explicitly.

The immediate-parent check is a security boundary: it prevents unrelated windows and nested frames from controlling the sketcher, while exact-origin replies avoid sending results with a wildcard target. It does not restrict which sites may embed the page; framing policy remains a deployment concern.

Switching from Asyncify to JSPI would not simplify the message protocol itself—the request correlation, origin checks, and replies remain necessary. The wrappers already await results, so they should fit either model, but JSPI changes the WASM/Embind build contract and browser requirements. I would handle that as a separate build migration.

d-b-w:
> Just curious what use case is this for??

I want to write a Jupyter plugin and sidestep CORS protections. I've tried to write a jupyter plugin that rehosts a sketcher instance, but it is very, very annoying.

test/wasm/html_api.test.js
```diff
@@ -0,0 +1,273 @@
...
+      },
+    },
+    document: { querySelector: () => ({ style: {}, addEventListener() {} }) },
+    console: { error() {} },
+    schrodinger_sketcher_entry() {},
+    qtLoad: loadModule,
+  });
+  vm.runInContext(script, context);
+  return context;
+}
```
juzerzarif: Is this like a homegrown JSDOM? Why not just run these tests in the browser with playwright? Do we need to use the sketcher in NodeJS somewhere?

PR #431, opened 2026-09-16
SKETCH-2857: Unbound attachment points are now based on Monomer DB and SMILES strings

KevKeating:
- Linked Case: SKETCH-2857, SKETCH-2834

### Description

Previously, unbound attachment points were determined somewhat generically.  For example, all PEPTIDE monomers got N and C attachment points, except for cysteine, which also had an S attachment point.  This led to several issues:
- some non-cysteine peptide monomers should also have a side chain attachment point, such as lysine
- non-native analogs and SMILES monomers were treated as generic peptides and only got N and C attachment points, regardless of their specific structure

This PR changes the logic for determining unbound attachment points so that it's always based on the SMILES string.  For monomers in the monomer database (including natives), the SMILES string is retrieved from the database.  Additionally, for the peptide side chain attachment point (R3), the display name of the attachment point is determined by examining the element: sulfurs are labeled as "S" and other elements are labeled as "X".  Note that if we always labeled side-chain attachment points with the element, there wouldn't be a way to distinguish a side-chain N attachment point from the N-terminal backbone attachment point.  We could certainly tweak things in the future to allow for that, though. 

A few additional notes:
- Side-chain interactions between two sulfurs are now drawn as disulfides, while other side-chain interactions are not.  Previously, disulfides were only drawn as such if they were between two cysteines.
- This change affects more than just peptide monomers; attachment points for nucleic acid monomers and CHEM monomers are handled the same way (i.e. check the SMILES string to determine whether the attachment point is present).
- For everything other than peptide side chains, this change only affects whether the attachment point is present or not.  The display name isn't affected at all.  For example, all peptide R1 attachment points are still labeled as "N", and all nucleic acid sugar R3 attachment points are still labeled as "C1'".

### Important note
As a result of this change, any SMILES monomers **must** have its attachment points labeled.  Otherwise, Sketcher will refuse to draw new connections to/from the monomer.

### Testing Done

I tested the new behavior on the Windows desktop app.  I also added new unit test coverage for the new behavior.  Some of the new tests were written by Codex, but I wrote a number of them by hand.

~Note that the new test involving a SMILES monomer requires the fix for SKETCH-2861.  Otherwise, the test would run into that bug and fail.  To avoid that, this PR is based on that PR's branch instead of `main`.  I'll switch this PR over to `main` once that PR is merged.~


src/schrodinger/sketcher/rdkit/monomeric.cpp
```diff
@@ -73,11 +78,139 @@ class NoAvailableDirectionsException : public std::exception
...
+            // parse the attachment point number after removing the leading
+            // underscore
+            const auto parsed_num = ap_name_to_num(atom_label.substr(1));
+            if (parsed_num > 0) {
+                attachment_point_num = parsed_num;
+            }
+        }
+        if (attachment_point_num == 0 && atom->getAtomicNum() == 0) {
+            attachment_point_num = atom->getIsotope();
+        }
```
ethan-schrodinger, line 175: >isotope-numbered dummy atoms

It might be worth adding some documentation here (or I guess in the header file) about where this isotope-numbered dummy atom convention comes from.

src/schrodinger/sketcher/rdkit/monomeric.cpp
```diff
@@ -73,11 +78,139 @@ class NoAvailableDirectionsException : public std::exception
...
+        atom->getPropIfPresent(RDKit::common_properties::atomLabel, atom_label);
+
+        if (!found_mol_atom_map_num && atom_label.starts_with("_R")) {
+            // parse the attachment point number after removing the leading
+            // underscore
+            const auto parsed_num = ap_name_to_num(atom_label.substr(1));
+            if (parsed_num > 0) {
+                attachment_point_num = parsed_num;
+            }
+        }
```
ethan-schrodinger, line 172: Would it make sense to move this string parsing into its own function?

Specifically: is this an intuitively distinct piece of logic that we'd get some benefit from documenting and testing separately, or is it not worth the trouble?

src/schrodinger/sketcher/rdkit/monomeric.cpp
```diff
@@ -73,11 +78,139 @@ class NoAvailableDirectionsException : public std::exception
...
+        if (attachment_point_num == 0 && atom->getAtomicNum() == 0) {
+            attachment_point_num = atom->getIsotope();
+        }
+        if (attachment_point_num == 0) {
+            continue;
+        }
```
ethan-schrodinger: What does this mean?

src/schrodinger/sketcher/rdkit/monomeric.cpp
```diff
@@ -73,11 +78,139 @@ class NoAvailableDirectionsException : public std::exception
...
+        const RDKit::Atom* heavy_atom = atom;
+        if (heavy_atom->getAtomicNum() <= 1) {
+            heavy_atom = nullptr;
+            for (const auto* neighbor : mol->atomNeighbors(atom)) {
+                if (neighbor->getAtomicNum() > 1) {
+                    heavy_atom = neighbor;
+                    break;
+                }
+            }
+        }
```
ethan-schrodinger, line 189: To what extent are these SMILES strings validated before they reach this function? Specifically, are we protected from weird things like multiple attachment points associated with the same attachment point number, or, e.g., dummy atoms that are not bound to a heavy atom? If we encounter such things, is it correct to just move on, or should we produce a warning or exception?

src/schrodinger/sketcher/rdkit/monomeric.cpp
KevKeating: I'm not sure if it comes from anywhere per se.  My understanding is that, since vanilla SMILES (i.e., not extended SMILES) doesn't have any official standard for numbering R-groups, people just shoehorned the numbers into anything that accepted numbers and isotopes happened to be one of those ways.

src/schrodinger/sketcher/rdkit/monomeric.cpp
KevKeating: Yeah, that's definitely reasonable.  On my local `SKETCH-2826` branch, I've moved this logic into a separate `get_attachment_point_num` function so it can be reused when we normalize the attachment point (i.e., make sure that attachment points show up as R-groups when we read a SMILES string into the custom monomer dialog).  I'll backport that change to here.

src/schrodinger/sketcher/rdkit/monomeric.cpp
KevKeating: I've been assuming that this function would only be used on SMILES strings that have already been sanity checked, and that anything that imports or modifies SMILES is responsible for validating SMILES before adding them to MolModel.  It's definitely worth documenting that assumption, though.  I'll add a note to the docstring for this method.

src/schrodinger/sketcher/tool/abstract_draw_monomer_or_monomeric_connection_scene_tool.cpp
```diff
@@ -196,19 +210,12 @@ UnboundMonomericAttachmentPointItem* get_default_attachment_point(
         UnboundMonomericAttachmentPointItem* default_ap = nullptr;
         if (tool_type == MonomerType::PEPTIDE) {
             std::vector<int> aps_to_look_for = {PeptideAP::C, PeptideAP::N};
-            if (hovered_res_name == CYS_RES_NAME &&
-                tool_res_name == CYS_RES_NAME) {
-                // only offer to form a disulfide if we're forming a connection
-                // between two cysteines
-                aps_to_look_for.push_back(PeptideAP::S);
```
ethan-schrodinger: It looks like previously we would check that both the hovered residue and the tool were cysteine. Now it's more general, in a sense -- we check that the hovered residue has AP3 -- but we don't check anything about the tool. Is this correct behavior?

ethan-schrodinger (approved):

src/schrodinger/sketcher/tool/abstract_draw_monomer_or_monomeric_connection_scene_tool.cpp
KevKeating: Yeah, there shouldn't be any change in behavior from this since the downstream logic is now smart enough to avoid forming R3-R3 interactions when one of the monomers lacks an R3.

PR #432, opened 2026-09-16
SKETCH-2861: Removing remove_cxsmiles_labels since it corrupts HELM strings

KevKeating:
- Linked Case: SKETCH-2861

### Description

The `remove_cxsmiles_labels` function replaces any HELM atom labels that are longer than 4 character with "CX".  This means that any SMILES strings or non-native monomer name of that length gets replaced.  As a result, pasting something like
```
PEPTIDE1{[O=C([C@H](CSCCN[H:3])N[H:1])[OH:2]]}$$$$V2.0
```
or
```
PEPTIDE1{[LONGNONNATIVE]}$$$$V2.0
```
into the Sketcher and then exporting it as a HELM string will give
```
PEPTIDE1{[CX]}$$$$V2.0
```
instead of anything resembling the input HELM string.  Exporting one of these structures as a SMILES string will instead crash the Sketcher since it attempts to parse "CX" as a SMILES string.  (Update: I don't think this will crash anymore since I just copied the commit for SHARED-12694 from mmshare over to this repo, but the export still won't work correctly without this PR.)

I'm not sure what purpose `remove_cxsmiles_labels` is intended to serve.  @rachelnwalker, do you know?  This PR just removes it.  Is that a reasonable thing to do?

### Testing Done

Added additional unit tests that would've caught these issues.


ethan-schrodinger (approved):
>I'm not sure what purpose remove_cxsmiles_labels is intended to serve. @rachelnwalker, do you know? This PR just removes it. Is that a reasonable thing to do?

My first thought is that this was just done for display purposes. Is there any sense in attempting to preserve the full HELM string while also showing some truncated version in contexts where we wouldn't necessarily want as much detail?

rachelnwalker (approved):
@KevKeating Thanks for doing this. I think this is an artifact from when monomeric image generation was a python script and we didn't want to deal with HELM strings rendering. This makes sense

KevKeating:
> My first thought is that this was just done for display purposes.  Is there any sense in attempting to preserve the full HELM string while also showing some truncated version in contexts where we wouldn't necessarily want as much detail?

Yeah, that's what will happen after this PR.  The truncating is done at the QGraphicsItem level, and the RDKit molecule preserves the full string.

> @KevKeating Thanks for doing this. I think this is an artifact from when monomeric image generation was a python script and we didn't want to deal with HELM strings rendering. This makes sense

Ahh, that makes sense!  I'd completely forgotten about the Python rendering!

PR #435, opened 2026-09-16
Compress Sketcher WebAssembly for npm consumers

JarrettSJohnson:
## Summary

- ship the npm package main Sketcher WebAssembly module as a deterministic gzip asset instead of the 44 MB raw file
- inject a gzip-first Emscripten loader into `Sketcher.js`, preserving cache-bust queries, `locateFile` overrides, streaming compilation, and the raw-asset fallback used by development and non-npm deployments
- add unit coverage for static gzip decoding, server-decoded gzip, raw fallback, consumer-provided instantiation, and fatal load failures

The current WASM artifact is 44,181,715 bytes raw and 13,226,330 bytes with `gzip -9 -n`, a reduction of about 70% for consumer transfers.

This follows the delivery approach used for WebPyMOL in:

- https://github.com/schrodinger/pymol-incentive/pull/761
- https://github.com/schrodinger/pymol-incentive/pull/762

## Testing

- `npm run test:unit` — 5 tests passed
- Emscripten `--pre-js` compiler smoke test
- `node --check wasm/compressed_wasm_loader.js`
- `git diff --check`
- parsed the modified GitHub Action as YAML


juzerzarif:
I don't know if this is something we need to hand wire in the loader ourselves. Most web assets are served gzipped but like JS bundlers don't really deal with the fetch compressed asset + decompress stuff themselves. 
You just tell nginx to gzip files and expect it to provide the correct headers to clue the browser into handling decompression. I think that's the layer where configuring this makes sense.

As a "library", I think it makes sense here to maybe provide pre-compressed assets as a convenience so consumers don't have to know which assets to compress and nginx doesn't need to dynamically compress the large wasm binary at request time.

PR #436, opened 2026-09-17
SKETCH-2786: Connections between sugars and phosphates now use the correct phosphate attachment point

KevKeating:
- Linked Case: SKETCH-2786

### Description

The two attachment points on a phosphate are chemically identical since it's just a single atom.  However, the HELM spec still specifies that standard backbone connections between a sugar and a phosphate should always be R2-R1 or R1-R2.  If a connection is R1-R1 or R2-R2, then it's considered a custom bond, even though there's no chemical difference between these connections and the standard ones.  This PR makes sure that Sketcher always creates a standard bond between a sugar and a phosphate, even if the user drags to or from the "wrong" phosphate attachment point.

### Testing Done

Codex added a unit test, and I tested the behavior on the Windows desktop app.  On `main`, intentionally dragging to the wrong phosphate attachment points generated the following HELM strings:
```
RNA1{R}|RNA2{P}$RNA1,RNA2,1:R2-1:R2$$$V2.0
RNA1{R}|RNA2{P}$RNA1,RNA2,1:R1-1:R1$$$V2.0
```
With this PR, the same actions generated:
```
RNA1{R.P}$$$$V2.0
RNA1{P.R}$$$$V2.0
```


test/schrodinger/sketcher/tool/test_draw_monomeric_connection_scene_tool_integration_tests.cpp
```diff
@@ -17,6 +18,66 @@ namespace schrodinger
...
+// SKETCH-2786: both drawing tools normalize only a phosphate with two free
+// backbone attachments, and only when connecting to sugar R1 or R2.
+BOOST_DATA_TEST_CASE(test_sugar_phosphate_attachment_selection,
+                     boost::unit_test::data::make({false, true}) *
+                         boost::unit_test::data::make({false, true}) *
+                         boost::unit_test::data::make({1, 2, 3}) *
+                         boost::unit_test::data::make({1, 2}) *
+                         boost::unit_test::data::make({0, 1, 2}),
+                     connection_tool, phosphate_first, sugar_ap, phosphate_ap,
+                     occupied_ap)
```
ethan-schrodinger, line 42: It may be worth explicitly commenting about what `phosphase_first` means and why it's important to include this in our test.

ethan-schrodinger (approved):
LGTM

test/schrodinger/sketcher/tool/test_draw_monomeric_connection_scene_tool_integration_tests.cpp
KevKeating: I expanded the docstring for this test:
```cpp
/**
 * Make sure that the monomer tool and the monomeric connection tool normalize
 * bonds between phosphates and sugars. This normalization should only happen
 * when connecting to/from a phosphate with both R1 and R2 available, and only
 * when connecting from/to the sugar's R1 or R2 attachment point.
 * @param connection_tool whether to use the monomeric connection tool or the
 * monomer tool
 * @param phosphate_first whether to mimic a drag from a phosphate to a sugar
 * (true) or from a sugar to a phosphate (false)
 * @param sugar_ap the sugar attachment point to use for the connection
 * @param phosphate_ap the phosphate attachment point to use for the connection
 * @param occupied_ap which (if any) of the phosphate's attachment points are
 * already involved in a connection
 */
```

PR #437, opened 2026-09-17
SKETCH-2862: Ensure that residue names exist in the monomer DB for imported monomeric structures and that all SMILES are valid

KevKeating:
- Linked Case: SKETCH-2862

### Description

This adds in a sanity check when importing molecules that all monomer names exist in the monomer database and that all SMILES strings are valid.  If not, the Sketcher will put up an error dialog.  Without this check, it's possible to paste something like
```
PEPTIDE1{[THIS_IS_NOT_A_REAL_RESIDUE_NAME]}$$$$V2.0
```
into the Sketcher, which then leads to crashes with the code from https://github.com/schrodinger/sketcher/pull/431 that uses the monomer database to determine how many attachment points to draw.  Because it fixes a crash, I'm assuming that we want to merge this into 26-4.

### Testing Done

Codex added new unit tests.  I sanity checked the behavior on the Windows desktop app:

image


ethan-schrodinger (approved):
FWIW, Codex flagged that if `addMol()` fails validation as part of `SketcherWidget::importText`, we'll clear the sketcher and then not end up actually importing anything. I'm not sure if we care.

KevKeating:
> FWIW, Codex flagged that if `addMol()` fails validation as part of `SketcherWidget::importText`, we'll clear the sketcher and then not end up actually importing anything. I'm not sure if we care.

Yeah, agreed.  That seems less than ideal, but probably not worth worrying _all_ that much about.  We have the same behavior in atomistic mode if a user pastes a SMILES string with more than 200 atoms, although an unrecognized HELM monomer admittedly seem more likely than a ridiculously long SMILES string.  I filed SKETCH-2866 for this problem, but we probably don't need to prioritize it unless someone complains.

PR #438, opened 2026-09-18
SKETCH-2741: Adding a custom monomer tool with a Sketcher dialog

KevKeating:
- Linked Case: SKETCH-2741

### Description

This PR adds a custom monomer tool with a separate dialog for the user to sketch and/or paste in their custom monomer.  The spec for this dialog isn't yet finalized so we might need to tweak some of the specifics of the bottom bar and the dialog's starting structure, but the overwhelming majority of the code here would remain the same.  A few notes:

- This new dialog is currently only hooked up for new monomers; there's still no way to edit an existing monomer in the dialog.  I'll implement that in a separate PR.
- This PR allows the user to mutate a monomer to a SMILES monomer even if the monomer has connections to an attachment point that isn't present in the SMILES.  It also allows the user to click-and-drag to create a dimer of SMILES monomers, even if the SMILES doesn't have any attachment points.  Fixing these issues will require some functions from https://github.com/schrodinger/sketcher/pull/431 plus a non-trivial amount of additional code, so I'll put up a separate PR for that.
- Currently, rdkit_extensions::mutateMonomer guesses as to whether `name` is a residue name or a SMILES string.  I've added an optional Boolean argument that allows the caller to explicitly specify that, but preserves the original auto-detection behavior by default.  @rachelnwalker, this is probably the only part of the PR that's relevant to you.
- The icon for the custom monomer button is a PNG that I clipped from Laura's spec.  We should ask Laura to make an SVG of the icon and switch to that, but that's probably not overly critical for 2026-4 if Laura's overloaded.
- SketcherWidget currently assumes that all custom nucleic acid monomers are bases since the dialog doesn't allow the user to distinguish between bases, sugars, and phosphates.  Fixing that will require a fix for SKETCH-2796.  Chris had suggested creating a new custom nucleic acid monomer type (e.g. `MonomerType::NA_CUSTOM`) to handle this, which seems reasonable to me.  Regardless of the solution we come up for that case, we can improve how SketcherWidget handles this scenario as part of that case as well.
- A few hundred of the changed lines in this PR are from ui files, so this review isn't _quite_ as large as it looks at first glance.

### Testing Done

Added unit tests and manually tested the new tool on the Windows desktop app.  I've also had Codex review this code a few times looking for edge cases and fixing what it finds.  In its most recent passes, it's only found expected issues (i.e. the things listed above) and problems caused by the HELM format itself (e.g. HELM doesn't have a way to explicitly distinguish between a residue name and a SMILES string as far as I know, so creating a monomer with a SMILES string that happens to match a residue name leads to downstream issues).


src/schrodinger/sketcher/dialog/custom_monomer_dialog.cpp
```diff
@@ -0,0 +1,97 @@
...
+    // between the SketcherWidget and the edge of the dialog
+    m_dlg_layout->setContentsMargins(0, 0, 0, 0);
+    qobject_cast<QVBoxLayout*>(layout())->setContentsMargins(0, 0, 0, 0);
+
+    ui->monomer_type_combo->addItem("Amino acid",
+                                    QVariant::fromValue(ChainType::PEPTIDE));
+    ui->monomer_type_combo->addItem("Nucleic acid",
+                                    QVariant::fromValue(ChainType::RNA));
+    ui->monomer_type_combo->addItem("CHEM",
+                                    QVariant::fromValue(ChainType::CHEM));
```
ethan-schrodinger, line 41: Laura is very much not on board with the combo box idea. For the time being, it may make sense to not have any widget here (neither combo box nor checkbox), and instead default to using context to decide what type of structure the monomer is (e.g., which palette is visible). This would obviously mean there is no way to create new SMILES-based CHEM monomers, but maybe that's okay for now.

include/schrodinger/rdkit_extensions/monomer_mol.h
```diff
@@ -93,18 +94,22 @@ RDKIT_EXTENSIONS_API std::unique_ptr<Monomer>
...
+ * string or a residue name. If not given, will be automatically determined
+ * based on helm_symbol.
  */
-RDKIT_EXTENSIONS_API void mutateMonomer(RDKit::ROMol& monomer_mol,
-                                        unsigned int monomer_idx,
-                                        std::string_view helm_symbol);
+RDKIT_EXTENSIONS_API void
+mutateMonomer(RDKit::ROMol& monomer_mol, const unsigned int monomer_idx,
+              const std::string_view helm_symbol,
+              const std::optional<bool> is_smiles_opt = std::nullopt);
```
rachelnwalker: I think this will require SWIG changes in mmshare - could we just name the option `is_smiles`?

rachelnwalker (commented):
LGTM for the `rdkit_extensions` changes.

include/schrodinger/rdkit_extensions/monomer_mol.h
KevKeating: Sure, that's no problem.  I made that change and confirmed that mmshare builds successfully after porting this PR over.

rachelnwalker (approved):
Approving `rdkit_extensions` changes

